PR #9, opened 2019-06-07
Add organisation switcher

whatterz:
Added:
- Styles
- Macro


gavinwye:
@adamsilver Valid point, but we can document that as a known issue and fix it. Once we have teams using a consistent switcher it will be easy to change it.

PR #10, opened 2019-06-07
Add currency input component

whatterz:
Added:
- Styles
- Macro

### DIFF
--- a/app/views/components/currency-input/index.html
+++ b/app/views/components/currency-input/index.html
@@ -1,0 +1,56 @@
+{% extends "layouts/base.html" %}
+
+{% block body %}
+
+  <div class="govuk-grid-row">
+
+    <div class="govuk-grid-column-full">
+
+      {{ govukBackLink({
+        text: "Back",
+        href: "../"
+      }) }}
+
+      <h1 class="govuk-heading-xl" tabindex="-1">
+        <span class="govuk-caption-xl">Components</span> Currency input
+      </h1>
+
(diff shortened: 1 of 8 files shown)


app/views/components/currency-input/index.html
```diff
@@ -0,0 +1,56 @@
...
+        }
+      }) }}
+
+      {{ mojCurrencyInput({
+        id: "amount",
+        classes: "govuk-input--width-10",
+        name: "amount",
+        currencyLabel: {
+          text: "&yen;",
+          label: "Yen",
```
adamsilver: I can't see where this is used in the codebase.

app/views/components/currency-input/index.html
```diff
@@ -0,0 +1,56 @@
...
+          text: "Amount",
+          classes: 'govuk-!-font-weight-bold'
+        },
+        hint: {
+          text: "Enter the amount you want to exchange"
+        }
+      }) }}
+
+      {{ mojCurrencyInput({
+        id: "amount",
```
adamsilver: this id conflicts with the input above and causes the label to break.

app/views/components/currency-input/index.html
whatterz: I have removed the `label` param

app/views/components/currency-input/index.html
whatterz: I have made the `id`s unique

PR #12, opened 2019-06-07
Add notification badge component

whatterz:
Added:
- Styles
- Macro

### DIFF
--- a/app/views/components/notification-badge/index.html
+++ b/app/views/components/notification-badge/index.html
@@ -1,0 +1,42 @@
+{% extends "layouts/base.html" %}
+
+{% block body %}
+
+  <div class="govuk-grid-row">
+
+    <div class="govuk-grid-column-one-half">
+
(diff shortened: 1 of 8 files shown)


adamsilver:
The demo has the same id on all of the examples. I think it's because you added a default `id` attribute. I don't think we should have a default id and instead just provide the attributes param as we talked about last week. This might be applicable to other components too.

I'm worried about the name and specificity of the component name. We already (will) have a badge component. Could we brainstorm some ideas for names and get the working group's help (over slack will do?).

Throwing these out there: Alert, Bubble, Counter for the moment.

whatterz:
@adamsilver I will remove the default `id` attribute from the component in a subsequent PR.

With respect to the name of the component, we will have `mojBadge` which relates to status and `mojNotificationBadge` which relates to the red-circle notifications found on many apps and also in the HMRC pattern library. We certainly need to reconcile names.

PR #13, opened 2019-06-07
Add header component

whatterz:
Added:
- Styles
- Macro


whatterz:
@adamsilver I think we need to do several things in the next iteration of the header:

- Replace the `organisation` object with `organisationLabel`
- Replace the `service` object with `serviceLabel`
- Replace the `navigation` object with `accountNavigation`
- Implement the `primaryNavigation` object, rather than use `call` as you recommend
- Implement crown or crest logo functionality

adamsilver:
Cool. Merge away :)

PR #15, opened 2019-06-07
Add add another component

whatterz:
Added:
- Styles
- Macro


app/views/components/add-another/index.html
```diff
@@ -0,0 +1,116 @@
...
+            }
+          },
+          hint: {
+            text: "For example, 31 3 1980"
+          },
+          items: [
+            {
+              id: "person[0][dob][day]",
+              name: "day",
+              classes: "govuk-input--width-2"
```
adamsilver: You need to add attributes onto the date inputs like we do for the fields above to make the labels accessible when cloning. The component is fine. Just the demo that needs sorting.

app/views/components/add-another/index.html
```diff
@@ -0,0 +1,116 @@
...
+      {{ govukBackLink({
+        text: "Back",
+        href: "../"
+      }) }}
+
+      <h1 class="govuk-heading-xl" tabindex="-1">
+        <span class="govuk-caption-xl">Components</span> Add another
+      </h1>
+
+      <h2 class="govuk-heading-l moj-add-another__heading">Add a person</h1>
```
adamsilver: This needs to be inside the component container for the component script to work. In this case . you have the component as the form below the heading.

app/views/components/add-another/index.html
whatterz: Done

PR #19, opened 2019-06-10
Update currency input documentation

whatterz:
Added the following:

- Links to guidance (MOJ Design System and MOJ Frontend)
- Links to dependencies (In this instance, to the GOV.UK Frontend hint, label and error message components)
- Examples (default example and one showing a different currency being injected)
- Attribute guidance (for the main container, currency, hint, label and error message objects)


gavinwye:
Did this come from HMCTS? Should we be name-checking them?

adamsilver:
Yep, this is from HMCTS. Name is consistent.

PR #22, opened 2019-06-10
Update sub navigation documentation

whatterz:
- Add component examples
- Add component argument definitions


gavinwye:
Would be good to show an example of the route that's needed to mack the nav switch state when you navigate around a service. That's missing from the GOV.UK design system at the moment and I haven't been able to get it to work.

If we don't do this now we could follow up quite quickly with this. I think it's more important to have it in and iterate than have it working perfectly.

PR #23, opened 2019-06-13
Add multi select component

adamsilver:


whatterz:
I have submitted a bug to be fixed at a later date, so that we can get this PR merged

https://github.com/ministryofjustice/mojdt-frontend/issues/30

PR #32, opened 2019-06-14
Add filter from HMCTS

adamsilver:
- Migrated filter across from HMCTS
- Added attributes param onto the container


adamsilver:
Done. Ready for you @whatterz

PR #38, opened 2019-06-20
Add timeline component

whatterz:
This is an evolution of HMCTS timeline.

Added:
- Styles
- Macro

Documentation to follow!

### DIFF
--- a/app/filters.js
+++ b/app/filters.js
@@ -1,0 +1,18 @@
+const dateFilter = require('nunjucks-date');
+
+module.exports = function (env) {
+  /**
+   * Instantiate object used to store the methods registered as a
+   * 'filter' (of the same name) within nunjucks. You can override
+   * gov.uk core filters by creating filter methods of the same name.
+   * @type {Object}
+   */
+  let filters = {}
+
+  filters.date = dateFilter;
+
+  /* ------------------------------------------------------------------
+    keep the following line to return your filters to the app
+  ------------------------------------------------------------------ */
+  return filters;
+}
--- a/app/filters.js
+++ b/app/filters.js
@@ -18,10 +18,10 @@ module.exports = function (env) {
 
     switch(type) {
       case "datetime":
-        str = dateFilter(timestamp, 'D MMMM gggg') + " at " + dateFilter(timestamp, 'h:mm A');
+        str = dateFilter(timestamp, 'D MMMM gggg') + " at " + dateFilter(timestamp, 'h:mma');
         break;
       case "shortdatetime":
-        str = dateFilter(timestamp, 'D MMM gggg') + " at " + dateFilter(timestamp, 'h:mm A');
+        str = dateFilter(timestamp, 'D MMM gggg') + " at " + dateFilter(timestamp, 'h:mma');
         break;
       case "date":
         str = dateFilter(timestamp, 'D MMMM gggg');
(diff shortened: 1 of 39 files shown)


app/filters.js
```diff
@@ -0,0 +1,46 @@
...
+
+  filters.date = dateFilter;
+
+  function mojDateFilter(timestamp, type) {
+
+    // default datetime to timestamp
+    let datetime = timestamp;
+
+    switch(type) {
+      case "longdatetime":
```
adamsilver: I'd suggest a terser default “datetime” and then prefix with “short” only for short date.

app/filters.js
```diff
@@ -0,0 +1,46 @@
...
+        datetime = dateFilter(timestamp, 'D MMMM gggg') + " at " + dateFilter(timestamp, 'h:mm A');
+        break;
+      case "shortdatetime":
+        datetime = dateFilter(timestamp, 'D MMM gggg') + " at " + dateFilter(timestamp, 'h:mm A');
+        break;
+      case "longdate":
```
adamsilver: Same as above, so “date”

app/filters.js
```diff
@@ -0,0 +1,46 @@
...
+        datetime = dateFilter(timestamp, 'D MMMM gggg');
+        break;
+      case "shortdate":
+        datetime = dateFilter(timestamp, 'D MMM gggg');
+        break;
+      case "time":
+        datetime = dateFilter(timestamp, 'h:mm A');
```
adamsilver: This isn't quite right as it will always put minutes in. So 9am should be 9am (not 9:00am)

app/filters.js
```diff
@@ -0,0 +1,46 @@
...
+module.exports = function (env) {
+  /**
+   * Instantiate object used to store the methods registered as a
+   * 'filter' (of the same name) within nunjucks. You can override
+   * gov.uk core filters by creating filter methods of the same name.
+   * @type {Object}
+   */
+  let filters = {}
+
+  filters.date = dateFilter;
```
adamsilver: Why are we exposing this as a filter?

app/filters.js
```diff
@@ -0,0 +1,46 @@
...
+        break;
+      case "time":
+        datetime = dateFilter(timestamp, 'h:mm A');
+        break;
+    }
+
+    return datetime;
+  }
+
+  filters.mojDate = mojDateFilter;
```
adamsilver: Change this to:

```
filters.mojDate = function(datetimestamp, type) { ... }
```

Rather than defining it above and assigning it.

app/filters.js
whatterz: The `date` filter is available for the custom `format` param, for example:
```
{%- if item.datetime.type %}
  {{- item.datetime.timestamp | mojDate(item.datetime.type) -}}
{% else %}
  {{- item.datetime.timestamp | date(item.datetime.format) -}}
{% endif -%}
```
If the `type` parameter is provided then `format` is ignored. (Much like the `text` versus `html` params). We favour `type` over `format` and should perhaps be defaulted to "datetime" so provision is not required.

app/filters.js
whatterz: In line with `date` and `shortdate`, `datetime` and `shortdatetime`, I will introduce `shorttime` (9 AM) alongside `time` (9:00 AM).

What is the guidance around "am" versus "AM"? Aside: Grammarly always corrects me to uppercase.

whatterz: Is there a need for a `longdatetime`, for example: `Tuesday, 21 January 2019 at 2:15 PM`?

adamsilver: The GOV.UK style recommends always 9am and not 9:00am. So we shouldn't provide that option I don't think. It's also lowercase “am”.

whatterz: For case working systems, it's necessary to be more accurate with time than hours. The provision of `time` and `shorttime` gives this flexibility and follows `date` and `datetime` nomenclature.

PR #43, opened 2019-06-21
Update timeline component and documentation

whatterz:


src/components/moj-timeline/README.md
```diff
@@ -3,7 +3,81 @@
 - [Guidance](https://mojdt-design-system.herokuapp.com/components/timeline)
 - [Preview](https://mojdt-frontend.herokuapp.com/components/timeline)
 
+## Dependencies
```
adamsilver: I don't think we need to specify any of this here. The dependency should be included in the package

whatterz: So in the `package/package.json` should include the following?
```
"dependencies": {
  "moment": "^2.22.2"
}
```

app/filters.js
```diff
@@ -1,4 +1,3 @@
-const moment = require('moment');
 
 module.exports = function (env) {
```
adamsilver: Don't need to pass in `env` I don't think?

app/filters.js
```diff
@@ -9,58 +8,6 @@ module.exports = function (env) {
    */
   let filters = {}
 
-  /* ------------------------------------------------------------------
-    date filter for use in Nunjucks
-    example: {{ params.date | date("DD/MM/YYYY") }}
-    outputs: 01/01/1970
-  ------------------------------------------------------------------ */
-  filters.date = function(timestamp, format) {
```
adamsilver: We shouldn't expose this. But I think you said you wanted to raise an issue and deal with this outside of this ticket?

src/components/moj-timeline/README.md
```diff
@@ -3,7 +3,25 @@
...
+
+- [Moment.js](https://momentjs.com/)
+
+### Installation
```
adamsilver: This is not specific to timeline, these filters are available at the package level and so should be provided in the main readme/installation guides if they exist. They will in the design system.

src/filters/all.js
```diff
@@ -0,0 +1,69 @@
+const moment = require('moment');
+
+module.exports = function (env) {
```
adamsilver: Don't need `env` I don't think

app/filters.js
whatterz: Yes, a logging an issue is preferable. We need to consider what formats are used, for example "DD/MM/YYYY" is not encapsulated, yet used.

src/components/moj-timeline/README.md
whatterz: I have removed the line mentioning `moment.js` but will leave the installation instructions in lieu of another place to put the information.

app/filters.js
whatterz: This is the default construct for `app/filters.js`. I will leave as is and we can review at another time. I will create a ticket.

src/filters/all.js
whatterz: Removed.

PR #51, opened 2019-07-01
Filter toggle and layout

adamsilver:
This makes the filter be toggleable with the toggle button working responsively as part of a table action bar.

### DIFF
--- a/app/views/components/filter-layout/index.html
+++ b/app/views/components/filter-layout/index.html
@@ -96,10 +96,10 @@
                     text: 'Type'
                   },
                   items: [{
-                    href: '/path/to/remove/this',
+                    href: '#',
                     text: 'Divorce'
                   }, {
-                    href: '/path/to/remove/this',
+                    href: '#',
                     text: 'Financial remedy'
                   }]
                 },
@@ -108,7 +108,7 @@
                     text: 'Status'
                   },
                   items: [{
-                    href: '/path/to/remove/this',
+                    href: '#',
                     text: 'Application drafted'
                   }]
                 }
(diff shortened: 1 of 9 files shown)


app/views/components/filter-layout/index.html
```diff
@@ -0,0 +1,355 @@
...
+                text: 'Clear filters'
+              },
+
+              categories: [
+                {
+                  heading: {
+                    text: 'Type'
+                  },
+                  items: [{
+                    href: '/path/to/remove/this',
```
whatterz: It's worth replacing `/path/to/remove/this` with a simple `#` so that the app doesn't error.

app/views/components/filter-layout/index.html
```diff
@@ -0,0 +1,355 @@
+{% extends "layouts/base.html" %}
+
+{% block body %}
+
```
whatterz: To be consistent with other app examples, wrap the body content with the appropriate divs. For example:

```
{% block body %}
  <div class="govuk-grid-row">
    <div class="govuk-grid-column-full">
      <!-- backlink go here -->
      <!-- heading go here -->
      <!-- examples go here -->
    </div>
  </div>
{% endblock %}
```

adamsilver: Why don't we just remove these extra divs from the other examples? They're not needed right.

whatterz: If we assume everything sits in a grid row and column, then they are needed. Either way, it would be easier adding them to this component, rather than removing them from numerous others. 

We can always reconcile the views at a later date to use the appropriate divs.

PR #60, opened 2019-07-15
Component messages

whatterz:
Add a component to display messages between two actors, for example: caseworkers and legal professionals.

Resolves following issue on backlog: 
https://github.com/ministryofjustice/mojdt-design-system-backlog/issues/14

_NB. it's worth creating a pattern to demonstrate the message component in situ with input._

### DIFF
--- a/src/components/moj-messages/README.md
+++ b/src/components/moj-messages/README.md
@@ -1,0 +1,85 @@
+# Messages
+
+- [Guidance](https://moj-design-system.herokuapp.com/components/messages)
+- [Preview](https://moj-frontend.herokuapp.com/components/messages)
+
+### Installation
+
+You will need to install the following code at the bottom of `server.js`, just above `module.exports = app;`
+
+```
+// Add filters from MOJ Frontend
+let mojFilters = require('./node_modules/@ministryofjustice/frontend/filters/all')();
+mojFilters = Object.assign(mojFilters);
+Object.keys(mojFilters).forEach(function (filterName) {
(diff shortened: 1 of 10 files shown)


src/components/moj-messages/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+## Example
+Below is a typical example of the timeline component in use.
+
+```
+{{ mojMessages({
+  items: [
+    {
+      id: 1,
```
adamsilver: Why is an ID needed?

whatterz: It's not. It's optional. But _if_ someone wanted to link to a particular message, however unlikely, they could do so using the ID.

src/components/moj-messages/template.njk
```diff
@@ -0,0 +1,30 @@
...
+
+      {%- if previous_timestamp != item.timestamp | mojDate('date') %}
+
+        <time class="moj-message-list__date" datetime="{{ item.timestamp | mojDate('date') }}">
+          {{- item.timestamp | mojDate('date') -}}
+        </time>
+
+      {% endif -%}
+
+      <div {%- if (item.id) %} id="message_{{ item.id }}"{% endif %} class="moj-message-item moj-message-item--{{ item.type }}">
```
adamsilver: How will the user know a new message has been added in real time? Is this a real time thing? Have you considered aria live regions?

whatterz: In its current incarnation, this isn't a realtime component. Having said that, there is no reason why it could be enhanced.

PR #67, opened 2019-08-07
 Update organisation switcher link logic

### LINKED ISSUE #66: Organisation switcher output link element when no link is supplied
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem
    * Followed all applicable steps in the debugging guide
    * Checked the FAQs on the message board for common solutions
    * Checked that your issue doesn't exist: https://github.com/ministryofjustice/mojdt-frontend/issues
    * Checked that there is not already a package that provides the described functionality

### Description

Currently if you use the [organisation switcher](https://moj-design-system.herokuapp.com/components/organisation-switcher) without supplying a `link` object the template will still output a link but with an empty href attribute.

### Steps to Reproduce

1. Call `mojOrganisationSwitcher` with no link property
```
{{ mojOrganisationSwitcher({
  text: 'HMP Brixton'
}) }}
```

**Expected behaviour:**

We have a use case to use the organisation switcher to display a users current location but if they only have access to one location they have no need to "switch" or see this action.

The template should cater for this and only conditionally show the link element if a link is supplied.

**Actual behaviour:**

Will display change link
image

And output link element with no href value
image

**Reproduces how often:** 100%

teneightfive:
### Identify the bug

#66 

### Description of the change

Only display the organisation switcher link if a link object is
passed to the component.

This prevents an empty link tag being output to the source if there
is no link destination.

This change still supports the fallback text as long as a link `href`
is supplied.

It also updates the example to show how the component will look
without a change link and with the fallback link text.

### Possible drawbacks

This removes the ability to add HTML inside the link element. I can't see why we would want to accept HTML inside this particular element though and it felt like this implementation may have been copied from other components that may have a specific need for it. Where possible we should be limiting the amount of locations input can be marked as safe and displayed as HTML.

## What it looks like now

With this configuration:
```
{{ mojOrganisationSwitcher({
  text: 'HMP Brixton'
}) }}
```

It will display:
image

Fixes #66

### DIFF
--- a/app/views/components/organisation-switcher/index.html
+++ b/app/views/components/organisation-switcher/index.html
@@ -23,16 +23,22 @@
           href: '#'
         }
       }) }}
-  
+
       <br>
-      
+
       <h2 class="govuk-heading-l">Example 2</h2>
       {{ mojOrganisationSwitcher({
         text: 'HMP Brixton, HMP/YOI Feltham, HMP/YOI Isis, HMP Pentonville, HMP Thameside, HMP Wandsworth, HMP Wormwood Scrubs',
         link: {
-          text: "Change prisons",
           href: '#'
         }
+      }) }}
+
+      <br>
+
+      <h2 class="govuk-heading-l">Example 3</h2>
+      {{ mojOrganisationSwitcher({
+        text: 'HMP Brixton'
       }) }}
 
     </div>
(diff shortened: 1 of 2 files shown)


adamsilver:
Thanks @teneightfive — will be ready whehnwe next publish to NPM.

PR #80, opened 2019-09-12
Use CSS to create arrows for pagination

MalcolmVonMoJ:
Changed Pagination to use CSS instead of SVG, as discussed on Slack


adamsilver:
Thanks @MalcolmVonMoJ — this is great and will go out on the next release.

PR #85, opened 2019-10-09
Output JavaScript as UMD

gregtyler:

N/A but [discussed on Slack](https://mojdt.slack.com/archives/CH5RUSB27/p1570200776007000)

### Description of the change

Currently the JavaScript module is exposed as a global variable, and may be used by including the `all.js` file and accessing `window.MOJFrontend`. This doesn't support use cases with bundlers, where the package needs to export something.

Fortunately, there is a JavaScript design pattern called "[Universal Module Definition](https://riptutorial.com/javascript/example/16339/universal-module-definition--umd-)" (UMD) which allows the module to support various different export styles depending on the context it's used in. It only adds a few extra lines to the file and is backwards compatible. It is also [currently used](https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/compile-assets.js#L166) in the GOV.UK Design System, so would align the two design systems.

I've added a Gulp plugin called `gulp-umd` which simply wraps the `all.js` file in the UMD template.

### Alternative designs

I considered using Rollup, as the GOV.UK Design System does, but this would require significant unecessary changes to the JavaScript code base (it would need to be fully modularised).

I also considered manually adding the UMD template to existing files, adding the prefix to `namespace.js` and the suffix to a new `suffix.js` file. However, this would make the two files independently invalid, because brackets opened in `namespace.js` would be closed in `suffix.js`.

### Possible drawbacks

Other than `all.js` being a few lines larger, I can't think of any.

I considered that this might cause variable scope problems, but because `GOVUKFrontend` is locally scoped with `var` I don't believe that this could be possible.

### Verification process

I verified that the `all.js` file can be used in the following ways:

- Included in a `<script>` tag and referred to with global `GOVUKFrontend` (what is currently supported)
- Imported via `var MOJFrontend = require('.../all.js');` in Webpack
- Imported via `import MOJFrontend from '.../all.js';` in Webpack
- Imported via `import { ButtonMenu } from '.../all.js';` in Webpack

I specifically tested all of these with the [Button menu](https://moj-design-system.herokuapp.com/components/button-menu) component, but there's no reason to believe other components would work differently since they're all included in the `all.js` file.

### Release notes

- JavaScript module exported with Universal Module Definition

### DIFF
--- a/gulp/build-javascript.js
+++ b/gulp/build-javascript.js
@@ -1,5 +1,6 @@
 const gulp = require('gulp');
 var concat = require('gulp-concat');
+var umd = require('gulp-umd');
 
 gulp.task('build:javascript', () => {
   return gulp.src([
@@ -8,5 +9,13 @@ gulp.task('build:javascript', () => {
       'src/moj/components/**/*.js'
     ])
     .pipe(concat('all.js'))
+    .pipe(umd({
+      exports: function() {
+        return 'MOJFrontend';
+      },
+      namespace: function() {
+        return 'MOJFrontend';
+      }
+    }))
     .pipe(gulp.dest('package/moj/'));
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -4190,6 +4190,35 @@
         }
       }
     },
+    "gulp-umd": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/gulp-umd/-/gulp-umd-2.0.0.tgz",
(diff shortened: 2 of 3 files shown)


nickcolley:
Hey Greg,

I'm Nick I work on GOV.UK Frontend, Adam Silver asked me to take a peak at this.

We're big fans of UMD for the reasons you've explained here so I think the general approach is sensible.

Here you can see how GOV.UK Frontend is configured to output in UMD: https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/compile-assets.js#L165

In the future my hope is that GOV.UK Frontend can support native ES Modules but that's a bit of a personal hope and not something under serious consideration yet.

 I don't have time to properly check if your implementation is correct, so I would recommend someone close to this project try out the different examples you've noted in your description.

Issue #92, opened 2020-05-18
New Cookie Banner design and Javascript module


johnnolan:
## Summary

Add a common component for the Cookie Consent Banner required by all public facing sites.

## Motivation

Every site being released at the moment has their own implementation. Ideally we need this to be done once in the design system and consumed by the front ends.

The Javascript is also potentially quite complicated due to multiple permissions. We should create a module for people to easily implement the functionality too.

## Describe alternatives you've considered

The alternative is to continue fragmenting the implementation across multiple projects.

There is also an existing implementation but it is deprecated and requires an enormous amount of JS to get working.

## Additional context

Ideally we want feedback from all departments in MoJ to make sure this component handles all functionality and design requirements for everyone.

theDustRoom:
I've added this to the MOJ Frontend Backlog

theDustRoom:
An example: https://components.publishing.service.gov.uk/component-guide/cookie_banner

reyrrep:
Service: Send money to someone in prison.
Updated cookies in December 2019, making consent active.
URL: https://www.gov.uk/send-prisoner-money
![Pay_now_by_debit_card_–_Send_money_to_someone_in_prison_–_GOV_UK](https://user-images.githubusercontent.com/24682124/82307671-94574180-99b8-11ea-8e50-65c7ec2bde11.png)

![Cookies_–_Send_money_to_someone_in_prison_–_GOV_UK](https://user-images.githubusercontent.com/24682124/82307722-a638e480-99b8-11ea-9e28-cd0ad0d1c3b3.png)

PR #97, opened 2020-05-22
Added cookie banner component

johnnolan:

Relates to issue #92 

### Description of the change

This change adds a new Cookie Banner component. 

It does not contain any javascript currently. This component is purely for design purposes so that different projects can align on look and feel.

A separate pull request can be done post this to add a dynamic javascript implementation.

### Alternative designs

Previous designs have had 2 primary buttons in the header. This uses 1 primary and 1 secondary as per the GDS rules so it doesn't cause confusion to the user.

Some existing designs have a darker background, this has been removed to help increase colour contrast in the design.

### Possible drawbacks

None I can think of

### Verification process

To test we have implemented the design on the new Use a LPA service. It also exists in the CICA website. The design was taken from there and classes amended to fit the MoJ namespace.

I have used navigation via keyboard and also ran Axe and Wave across the component. All pass.

### Release notes

Added new Cookie Consent Banner component

### DIFF
--- a/app/views/components/cookie-banner/index.html
+++ b/app/views/components/cookie-banner/index.html
@@ -21,14 +21,18 @@
           title: 'Tell us whether you accept cookies',
           html: 'GOV.UK uses cookies which are essential for the site to work. We also use non-essential cookies to help us improve government digital services. Any data collected is anonymised.'
         },
-        acceptCookies: {
-          type: 'submit',
-          text: 'Accept all cookies'
-        },
-        setCookies: {
-          href: '/cookies',
-          text: 'Set application cookies'
-        }
+        items: [
+          {
+            type: 'submit',
+            text: 'Accept all cookies',
+            element: 'button'
+          },
+          {
+            href: '/cookies',
+            text: 'Set application cookies',
+            element: 'a',
+            class: 'govuk-button--secondary'
+          }]
       }) }}
 
   </div>
--- a/app/views/components/cookie-banner/index.html
+++ b/app/views/components/cookie-banner/index.html
@@ -16,6 +16,7 @@
       </h1>
 
       {{ mojCookieBanner({
+        show: true,
         heading: {
           title: 'Tell us whether you accept cookies',
           html: 'GOV.UK uses cookies which are essential for the site to work. We also use non-essential cookies to help us improve government digital services. Any data collected is anonymised.'
--- a/app/views/components/cookie-banner/index.html
+++ b/app/views/components/cookie-banner/index.html
@@ -1,0 +1,37 @@
+{% extends "layouts/base.html" %}
+
+{% block body %}
+
+  <div class="govuk-grid-row">
+
+    <div class="govuk-grid-column-full">
+
+      {{ govukBackLink({
+        text: "Back",
+        href: "../"
+      }) }}
+
+      <h1 class="govuk-heading-xl">
+        <span class="govuk-caption-xl">Components</span> Cookie banner
+      </h1>
+
+      {{ mojCookieBanner({
+        heading: {
+          title: 'Tell us whether you accept cookies',
+          html: 'GOV.UK uses cookies which are essential for the site to work. We also use non-essential cookies to help us improve government digital services. Any data collected is anonymised.'
+        },
+        acceptCookies: {
+          type: 'submit',
+          text: 'Accept all cookies'
+        },
+        setCookies: {
+          href: '/cookies',
+          text: 'Set application cookies'
+        }
+      }) }}
+
+  </div>
+
+</div>
+
+{% endblock %}
--- a/src/moj/components/cookie-banner/_cookie-banner.scss
+++ b/src/moj/components/cookie-banner/_cookie-banner.scss
@@ -8,7 +8,7 @@
   padding-bottom: govuk-spacing(3);
   left: govuk-spacing(3);
   padding-right: govuk-spacing(3);
-  background-color: govuk-colour("light-grey");
+  background-color: govuk-colour("white");
 
   &--show {
     display: block !important;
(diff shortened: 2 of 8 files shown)


app/views/components/cookie-banner/index.html
```diff
@@ -0,0 +1,37 @@
...
+          html: 'GOV.UK uses cookies which are essential for the site to work. We also use non-essential cookies to help us improve government digital services. Any data collected is anonymised.'
+        },
+        acceptCookies: {
+          type: 'submit',
+          text: 'Accept all cookies'
+        },
+        setCookies: {
+          href: '/cookies',
+          text: 'Set application cookies'
+        }
```
gregtyler, line 30: I think these arguments are too limiting: you can only use two buttons with pre-defined purposes, one is a button the other a link. I'd rather flexibility (so you can add "reject all", or give them both JS events) and a good copy-pastable example.

I think I'd prefer if this was an array of button configurations that are passed to GDS's [button component](https://design-system.service.gov.uk/components/button/). e.g.:

```suggestion
        [
            {
                value: 'Accept all cookies',
            },
            {
                href: '/cookies',
                text: 'Set application cookies'
            }
        ]
```

src/moj/components/cookie-banner/_cookie-banner.scss
```diff
@@ -0,0 +1,36 @@
...
+
+  box-sizing: border-box;
+
+  padding-top: govuk-spacing(3);
+  padding-bottom: govuk-spacing(3);
+  left: govuk-spacing(3);
+  padding-right: govuk-spacing(3);
+  background-color: lighten(
+    desaturate(govuk-colour("light-blue"), 8.46),
+    42.55
```
gregtyler, line 13: I'm not a fan of these magic numbers inside the component. Think we should either use an existing govuk colour, or [add a new one to the MOJDS](https://github.com/ministryofjustice/moj-frontend/blob/master/src/moj/settings/_colours.scss) so that it's formally defined.

src/moj/components/cookie-banner/template.njk
```diff
@@ -0,0 +1,24 @@
+{%- from "govuk/components/button/macro.njk" import govukButton %}
+
+<div class="cookie-banner govuk-width-container cookie-banner--visible" role="region" aria-label="cookie banner">
```
gregtyler, line 3:
```suggestion
<div class="cookie-banner govuk-width-container cookie-banner--show" role="region" aria-label="cookie banner">
```

I think the class should be `cookie-banner--show`? Was also wondering if this should be configurable by an argument?

gregtyler:
```suggestion
<div class="moj-cookie-banner govuk-width-container moj-cookie-banner--show" role="region" aria-label="cookie banner">
```
Ah, also all the classes need the `moj-` prefix.

src/moj/components/cookie-banner/template.njk
```diff
@@ -0,0 +1,24 @@
...
+    <div class="govuk-grid-row">
+        <div class=" govuk-grid-column-two-thirds">
+            <div class="cookie-banner__message">
+                <h2 class="govuk-heading-m">{{ params.heading.title }}</h2>
+                <p class="govuk-body">{{ params.heading.html }}</p>
```
gregtyler, line 8: Bit of an alignment with GOV.UK DS: allows `text` or `html` and allows `html` to have include HTML tags.
```suggestion
                <p class="govuk-body">{{ params.heading.html | safe if params.heading.html else params.heading.text }}</p>
```

src/moj/components/cookie-banner/_cookie-banner.scss
johnnolan: Good call, have changed this to the GDS light-grey standard

src/moj/components/cookie-banner/template.njk
johnnolan: Great catch, thanks, these have been updated

src/moj/components/cookie-banner/template.njk
johnnolan: Added!

app/views/components/cookie-banner/index.html
johnnolan: Yeah I like this idea, a lot more flexible, I will add this in

johnnolan: So in there we need the option of rendering a `button` or a `a`. In nunjucks, is it possible to specify nicely one or the other? Would you pass in a type property and render the appropriate element?

gregtyler: If you're using the GDS button component, you can pass in `element`. But it also automatically infers it from whether `value`(button) or `text`/`html`(a) was provided.([Button Nunjucks options](https://design-system.service.gov.uk/components/button/#options-button-example))

src/moj/components/cookie-banner/template.njk
```diff
@@ -0,0 +1,24 @@
...
+                <p class="govuk-body">{{ params.heading.html | safe if params.heading.html else params.heading.text }}</p>
+            </div>
+            <div class="moj-cookie-banner__buttons">
+                {%- for item in params.items %}
+                    <div class="govuk-grid-column-full govuk-grid-column-one-half-from-desktop">
+                        {{ govukButton({
+                            text: item.text,
+                            classes: 'govuk-button button--inline ' + item.class,
+                            type: item.type,
+                            element: item.element
```
gregtyler, line 17: This at least needs to forward `href` too
```suggestion
                            text: item.text,
                            classes: 'govuk-button button--inline ' + item.class,
                            type: item.type,
                            element: item.element,
                            href: item.href
```
It would be nicer if we could spread operator everything into `govukButton` but I don't think Nunjucks supports that.

gregtyler:
Approved from a technical perspective, probably someone else needs to approve the design?

PR #103, opened 2020-07-24
Issue 101 multifileupload does not work in ie11

steverendell:
Identify the bug: multifileupload does not work in ie11

Description of the change: removed use of unsupported FormData.get() method

Alternative designs n/a

Possible drawbacks n/a

Verification process: Browse component from test server using IE

Release notes n/a


src/moj/components/multi-file-upload/multi-file-upload.js
```diff
@@ -115,7 +115,7 @@ if(MOJFrontend.dragAndDropSupported() && MOJFrontend.formDataSupported() && MOJF
   MOJFrontend.MultiFileUpload.prototype.uploadFile = function(file) {
     var formData = new FormData();
     formData.append('documents', file);
-    var item = $(this.getFileRowHtml(formData.get('documents')));
+    var item = $(this.getFileRowHtml(file));
```
johnnolan, line 118: If this works, do we need formData defined at all?

steverendell: I dont think so - formData is passed to the ajax

johnnolan (approved):

PR #131, opened 2021-01-06
Create MOJFrontend.initAll()

gregtyler:

N/A. I thought this hard to discuss without having code to review, so intended this PR to be the discussion space.

#### Background

Several components require additional JavaScript to function properly (this isn't documented, something which I intend to remedy after this). For example, the [example password reveal component](https://moj-design-system.herokuapp.com/components/password-reveal) should contain some JavaScript too:
```html
<input class="govuk-input moj-password-reveal__input govuk-input--width-20" id="password" name="password" type="password" value="1234ABC!">

<script>
new MOJFrontend.PasswordReveal(document.getElementById('password'));
</script>
```

This isn't the case in the GOV.UK Design System. For example, take the similar [character count component](https://design-system.service.gov.uk/components/character-count/). Instead of manual JavaScript, this requires the attributes `data-module="govuk-character-count" data-maxlength="200"` and is automatically initialised alongside _all other_ components with the one-line call `GOVUKFrontend.initAll()`.

This approach is much easier to write, particularly for those unfamiliar with JavaScript. It's particularly helpful when using the Nunjucks templates (e.g. with the prototype kit) where `data-module` attributes can be added automatically and the author doesn't even need to consider JavaScript.

#### This change

This PR starts to add the same functionality to the MOJ Design System:  a global `initAll` function which initialises _some_ (see below) components automatically based on `data-module` attributes.

Six components are initially targeted:
- Add another
- Multi select
- Password reveal
- Rich text editor
- Search toggle
- Sortable table

I chose those six because they were easy to convert to an attribute-based pattern, both because of their limited configuration options and the structure of the underlying code.

Note that I also haven't made any notable changes to the underlying JavaScript code for any component, so existing manual-initialisation techniques will continue to work.

#### What wasn't included

I didn't touch any components which have no related JavaScript code. I also didn't include the following components because their JavaScript code is not trivial to convert. We would likely have to rewrite the components to support this initialisation approach (or at least add a likely-breaking-change to their signature):

- Button menu (has three optional configuration options)
- Filter toggle button (has a large number of configuration options)
- Form validator (defines bespoke validation rules, would likely benefit from a redesign based on HTML5 validation)
- Multi-file upload (has two required and three optional configuration options, two of which are URLs)

### Alternative designs

A much more light-touch approach would just be documenting the JavaScript each component needs, but that would still require authors to manually pen JavaScript and have at least basic knowledge of JS and—currently—jQuery.

There's also the option to commit more heavily and add this pattern for every component, but I believe that's beyond the scope of one PR.

### Possible drawbacks

If authors have code to manually initialise components (e.g. `new MOJFrontend.PasswordReveal(...);`) and then _also_ start using `MOJFrontend.initAll()`, some components will be initialised twice which may cause problems.

There's also the drawback noted above that some components still require JavaScript initialisation, which may be confusing. But clearer documentation can help with this.

### Verification process

- I ensured the code worked by adding `MOJFrontend.initAll()` to all the example pages in this repo. I also removed the bespoke JavaScript from the examples of components I've adapted.
- I ensured old functionality continued to work by adding this as a new layer rather than changing the underlying code. The only change to existing JavaScript is using `$($container)` to accept both jQuery and DOM objects as an argument to multi-selects.

### Release notes

- Added `MOJFrontend.initAll()` to automatically initialise several components if they have suitable data-* attributes

### DIFF
--- a/app/views/components/multi-select/index.html
+++ b/app/views/components/multi-select/index.html
@@ -15,10 +15,10 @@
         <span class="govuk-caption-xl">Components</span> Multi-select
       </h1>
 
-      <table class="govuk-table">
+      <table class="govuk-table" data-module="moj-multi-select" data-multi-select-checkbox="#select-all">
         <thead class="govuk-table__head">
           <tr class="govuk-table__row">
-            <th class="govuk-table__header moj-multi-select__select-all-container" scope="col"></th>
+            <th class="govuk-table__header" scope="col" id="select-all"></th>
             <th class="govuk-table__header" scope="col">Name</th>
             <th class="govuk-table__header" scope="col">Elevation</th>
             <th class="govuk-table__header" scope="col">Continent</th>
@@ -130,15 +130,4 @@
 
     </div>
   </div>
-{% endblock %}
-
-{% block pageScripts %}
-
-  <script>
-    new MOJFrontend.MultiSelect({
-      container: $('.moj-multi-select__select-all-container'),
-      checkboxes: $('.govuk-checkboxes__input')
-    });
-  </script>
-
 {% endblock %}
--- a/app/views/components/add-another/index.html
+++ b/app/views/components/add-another/index.html
@@ -11,7 +11,7 @@
         href: "../"
       }) }}
 
-      <div class="moj-add-another">
+      <div data-module="moj-add-another">
 
         <h1 class="govuk-heading-xl">
           <span class="govuk-caption-xl">Components</span> Add another
@@ -120,13 +120,5 @@
     </div>
 
   </div>
-
-{% endblock %}
-
-{% block pageScripts %}
-
-  <script>
-    new MOJFrontend.AddAnother($('.moj-add-another'));
-  </script>
 
 {% endblock %}
--- a/app/views/components/password-reveal/index.html
+++ b/app/views/components/password-reveal/index.html
@@ -48,11 +48,3 @@
   </div>
 
 {% endblock %}
-
-{% block pageScripts %}
-
-  <script>
-    new MOJFrontend.PasswordReveal(document.getElementById('password'));
-  </script>
-
-{% endblock %}
--- a/app/views/components/password-reveal/index.html
+++ b/app/views/components/password-reveal/index.html
@@ -33,7 +33,10 @@
           text: "Password",
           classes: 'govuk-label--m'
         },
-        value: "1234ABC!"
+        value: "1234ABC!",
+        attributes: {
+          'data-module': 'moj-password-reveal'
+        }
       }) }}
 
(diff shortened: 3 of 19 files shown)


app/views/components/multi-select/index.html
```diff
@@ -15,10 +15,10 @@ <h1 class="govuk-heading-xl">
         <span class="govuk-caption-xl">Components</span> Multi-select
       </h1>
 
-      <table class="govuk-table">
+      <table class="govuk-table" data-module="moj-multi-select" data-multi-select-checkbox="#select-all">
         <thead class="govuk-table__head">
           <tr class="govuk-table__row">
-            <th class="govuk-table__header moj-multi-select__select-all-container" scope="col"></th>
+            <th class="govuk-table__header" scope="col" id="select-all"></th>
```
johnnolan, line 21: Will this cause HTML Validation issues if there are multiple tables with the same ID? Would it be better as a default first column or override with optional defined property?

gregtyler: The selector is kind of irrelevant because it's just whatever's targetted by `data-multi-select-checkbox` on line 18. I agree that ID is a bad example though, and can change it to a data-* attribute.

I didn't want to make any assumptions about where the column would be to make sure we don't break any production sites. Insisting you define and specify one felt like the most clear and flexible rule.

johnnolan: Yeah that makes sense, nice one, all looking good to me

townxelliot:
This looks like a great idea. Is there any scope for annotating a component with "inited" or something to prevent the double initialisation issue?

gregtyler:
@townxelliot Thanks for the feedback! I've added an "is initialised" data property to the components affected by `initAll` now. If that's already set, a second attempt to initialise them will fail. We shouldn't need to worry about crossing over with the main design system, because we have different component names. I've prefixed all the data properties so that you can initialise two components on one element e.g. a table which is both sortable table and multi-select, which should also help with compatibility.

I'm not too keen on constraining `initAll`, because it reintroduces having to do manual configuration. You either call `initAll` in a shared file and need to list all the components you're using across your service (which potentially doesn't save much client work) or have to call it on every page, which would quickly become tricky to manage. I also think that the relative overhead of some `document.querySelector` calls is trivial, but we could add optional parameters later if it's problematic.

And yep, I'd hope the manual configuration to be temporary, but some components will need a bit of rewrite to work nicely without a block of JavaScript. Once we've deployed this PR I'll remove the any JS from the docs (though some components, like [Password reveal](https://moj-design-system.herokuapp.com/components/password-reveal), already don't mention it).

johnnolan (approved):

PR #163, opened 2021-06-10
chore: update various dependencies

gregtyler:
This fixes a few vulnerabilities in our toolchain. We wouldn't necessarily be affected by any of these, but it's best to be safe.


kelliedesigner (approved):
LGTM

PR #164, opened 2021-06-15
Fix syntax error

gregtyler:
I accidentally missed a quote mark and a comma, meaning the docs cannot render


BNewing (approved):
Looks good to me!

PR #168, opened 2021-06-16
docs: add last updated date to components

gregtyler:
Uses a new "lastUpdated" filter to automatically calculate the point at which the component's
source code was last changed (by analysing Git logs). Also provides a link to that change.

Also forces the docs to rebuilt on *every* push to main, so that changes to components are
immediately reflected in the last updated date.

<details>
<summary>Example</summary>

image

</details>

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -7,6 +7,7 @@ const matter = require("gray-matter");
 const mojFilters = require("./src/moj/filters/all");
 const nunjucks = require("nunjucks");
 const path = require("path");
+const { execSync } = require("child_process");
 
 module.exports = function (eleventyConfig) {
   const nunjucksEnv = nunjucks.configure([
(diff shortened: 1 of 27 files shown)


.eleventy.js
```diff
@@ -78,6 +79,15 @@ module.exports = function (eleventyConfig) {
     });
   });
 
+  eleventyConfig.addShortcode("lastUpdated", function (component) {
+    const dirPath = path.join(__dirname, "src/moj/components", component);
+    const lastUpdated = execSync(
+      `LANG=en_GB git log -n1 --pretty=format:%ad --date=format:'%e %B %Y' ${dirPath}`
```
gregtyler, line 85: Fun fact: I had to specify the language because otherwise it does it in your system's language 🇪🇸

johnnolan (approved):

PR #173, opened 2021-06-23
Add retired components section

kelliedesigner:
### Description of the change

- Remove Currency input and Tag components from component page navigation
- Add notification banners to Currency input and Tag page to direct to GOV.UK Design System components
- Add Retired components page

|Before |After  |
--- | ---
|ministryofjustice github io_moj-frontend_components_currency-input_ (1) | localhost_8080_components_currency-input_
|Component used to be accessed from the left navigation |Component is no longer accessed from the left navigation, and has a notification banner directing to the GOV.UK Design System |

|New retired components page |
--- |
|localhost_8080_components_retired-components_(iPad Pro) (1)|

### Release notes

Users will no longer see duplicate components in the MOJ Pattern Library which are also in the GOV.UK Design System, and can access retired components to find out what GOV.UK Design System component they should use.


gregtyler (approved):
This looks fantastic, nice one! 💃🏽

PR #175, opened 2021-06-24
Add component categories

kelliedesigner:
### Description of the change

- Group components in the navigation into four categories: Structure, Navigation, Forms, Content 
- Order components within categories based on how they may be ordered on the page where possible, instead of alphabetically

**Also in this PR**
- Add notification banner to the Badge component to communicate there is a similar component (Tag) in the GOV.UK Design System which should be considered. 

|Before |After  |
--- | ---
|ministryofjustice github io_moj-frontend_components_(iPad Pro) (1) |localhost_8080_components_(iPad Pro) (1) |
|ministryofjustice github io_moj-frontend_components_badge_ |localhost_8080_components_badge_ |

### Release notes

Users will be able to find components easier by browsing the categories, and seeing them listed as they would be ordered on the page.

### DIFF
--- a/docs/_includes/layouts/component.njk
+++ b/docs/_includes/layouts/component.njk
@@ -13,101 +13,130 @@
     <div class="govuk-grid-column-one-third">
       {{ appSideNavigation({
         classes: 'govuk-!-padding-top-6',
-        items: [
+        sections: [
           {
-            text: 'Add another',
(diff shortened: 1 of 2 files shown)


michaelchambersjnr (approved):
Looks good, thanks Kellie.

PR #178, opened 2021-06-25
Update notification badge documentation

kelliedesigner:
### Description of the change

- Remove duplicate guidance on Notification banner component which exists on HMRC Design Patterns
- Add banner with link to HMRC Design Patterns and the GOV.UK Design System backlog

### Further information

It was felt that the previous iteration using the structured headings taken from the GOV.UK Design System made the component look more 'finished' than it is, and didn't clearly reflect that there is only a small amount of guidance on the component to date. 

To better reflect the component as a work in progress in order to encourage contribution, and to reduce the overhead of maintaining guidance that exists elsewhere, it was agreed to remove duplicate guidance and reduce the page to the minimal information. 

The method of contribution will be added in the future.

|Before |After  |
--- | ---
|ministryofjustice github io_moj-frontend_components_notification-badge_(iPad Pro) (2) |localhost_8080_components_notification-badge_(iPad Pro) (2) |

### Release notes

Users will be able to see the how developed the Notification badge component and guidance is, and read further information on HMRC Design Patterns and GOV.UK.

### DIFF
--- a/docs/components/notification-badge.md
+++ b/docs/components/notification-badge.md
@@ -10,12 +10,12 @@ This component is in the GOV.UK Design System [community backlog](https://design
 {% endbanner %}
 
 {% example "/examples/notification-badge", 125 %}
-### When to use
+## When to use
 
 The notification badge lets the user know that there is new information to view, like unread messages, and how many of them there are. 
 
 Only use it if the number changes when the user performs an action.
-### When not to use
+## When not to use
 
 Do not use the notification badge when:
 
@@ -24,12 +24,12 @@ Do not use the notification badge when:
 
 Unless there is a strong user need, only use it as a part of the navigation.
 
-### How to use
+## How to use
 
 Display the notification badge to the right-hand side of the information it refers to.
 
 If the number is more than 99, display ‘99+’.
-### Research
+## Research
 
 Research shows that notification badges are common across online services, smartphones and apps. Usability testing showed:
 
--- a/docs/components/notification-badge.md
+++ b/docs/components/notification-badge.md
@@ -3,28 +3,12 @@ layout: layouts/component.njk
 title: Notification badge
 ---
 
-Use the notification badge component to notify users of something like an unread message.
+{% banner "HMRC Design Patterns has a similar component" %}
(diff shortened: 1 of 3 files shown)


gregtyler:
I'm a little bit dubious about this, since users might get confused that the HMRC DS page has different examples and different installation instructions. But I'm happy to give it a try.

I'm not a fan of adding the date though, since it doesn't explain what it represents. I have [a WIP PR](https://github.com/ministryofjustice/moj-frontend/pull/168) to add automatic dates to every component, which I think is slightly clearer (since it says "last updated") and has the benefit of auto-updating whenever changes are made.

kelliedesigner:
Comments should now be addressed, additionally there was a discussion about the 'How it works/How to use' section. 

> This section is a combination of:
> - How to install/use the component itself, ie copy the HTML or Nunjucks
> - How to use the component from a UI/design pov, ie info on positioning or behaviour.

This seems repetitive and increases cognitive load, as users may think it will be guidance on the specific component when it turns out to be the same text that on all components.

It was decided to include the installation instructions on the components landing page, instead of at the component level.

localhost_8080_components_

docs/components/notification-badge.md
```diff
@@ -3,28 +3,35 @@ layout: layouts/component.njk
...
+{% banner "HMRC Design Patterns has a similar component" %}
+[Notification badge](https://design.tax.service.gov.uk/hmrc-design-patterns/notification-badge/) in HMRC Design Patterns has a similar function and visual design to this component, and includes some guidance on when to use it.
 
-{% example "/examples/notification-badge", 125 %}
+This component is in the GOV.UK Design System [community backlog](https://design-system.service.gov.uk/community/backlog/) for review. 
+{% endbanner %}
 
-## When to use this component
+{% example "/examples/notification-badge", 125 %}
+### When to use
```
gregtyler, line 13: Sorry, a fiddly thing but all these headings should be at second level (i.e. `###` rather than `##`) otherwise screenreaders can't parse the page properly

gregtyler (approved):

PR #186, opened 2021-06-25
fix(add-another): find labels of affixed text inputs

### LINKED ISSUE #185: MOJ - Add another functionality doesn't work when input wrapper for suffix is in use
### Prerequisites

- [x ] Put an X between the brackets on this line if you have done all of the following:
  - Reproduced the problem
  - Followed all applicable steps in the debugging guide
  - Checked the FAQs on the message board for common solutions
  - Checked that your issue doesn't exist: https://github.com/ministryofjustice/moj-frontend/issues
  - Checked that there is not already a package that provides the described functionality

### Description

Currently we are looking at the moj-add-another (https://moj-design-system.herokuapp.com/components/add-another) design pattern and we have managed to get it working fine. However, when we wrap our input in a gov-uk wrapper (suffix) the add another button decides it no longer wants to work, interestingly it actually works when javascript is turned off.

### Steps to Reproduce

https://github.com/dwp/atw-payment-claims-service - link to prototype code in github. Specific folder structure is alpha>support-worker>hours-for-day.

**Expected behaviour:** [We expect that the add another functionality should still work when a govuk - suffix is added to the pattern for one of the fields as it would without]

**Actual behaviour:** [the grey add another button now is unresponsive and doesn't auto generate when the wrapper is in the code.]

**Reproduces how often:** [100%]

### Versions

You can get this information from the CHANGELOG. Also, please include details of the browser and OS version you're running.

### Additional Information
Apologies if this is not detailed enough, this is my first time leaving a bug issue on github and i'm a UX designer not a Developer, if you require any more information feel free to reach out to me on - owen.bennett@engineering.digital.dwp.gov.uk

gregtyler:
### Identify the bug

This was identified by @owen-bennett123 in #185

When a text input has affixes, it is wrapped in a `.govuk-input__wrapper` div, meaning its label is
no longer adjacent. This caused the Add anotehr component to fail, because it couldn't find the
input's label and update its for/id link.

### Description of the change

I've added a new strategy for finding the label: if it's not the parent or adjacent, then JavaScript
identifies it by searching for a label whose `for` attribute is the ID of the original input.

This should not introduce any regressions because the previous strategies are still in place and
prioritised: the new strategy will only be used if nothing was found, which would have previously
resulted in the application crashing.

I also did some cleanup:
 - Removed an unnecessary `script.js` from the Add another example
 - Restored the README for the Add another component (accidentally removed in a previous commit)

### Alternative designs

- I considered looking at the parent `.govuk-form-group` and finding the label inside it, but would cause issues with components that have multiple labels (like checkboxes or date inputs)
- I considered _only_ using my new strategy, rather than bothering to look for parent or adjacent labels, but that would have been a breaking change

### Verification process

I manually edited the [Add another example](https://design-patterns.service.justice.gov.uk/examples/add-another/), adding a prefix to the text inputs. I also copied the original example from the DWP repo in #185.

### Release notes

- fix(add-another): find labels of affixed text inputs

fixes #185

### DIFF
--- a/package/moj/components/add-another/add-another.js
+++ b/package/moj/components/add-another/add-another.js
@@ -42,9 +42,13 @@ MOJFrontend.AddAnother.prototype.getNewItem = function() {
 
 MOJFrontend.AddAnother.prototype.updateAttributes = function(index, item) {
 	item.find('[data-name]').each(function(i, el) {
+    var originalId = el.id
+
 		el.name = $(el).attr('data-name').replace(/%index%/, index);
 		el.id = $(el).attr('data-id').replace(/%index%/, index);
-		($(el).siblings('label')[0] || $(el).parents('label')[0]).htmlFor = el.id;
+
+    var label = $(el).siblings('label')[0] || $(el).parents('label')[0] || item.find('[for="' + originalId + '"]')[0];
+		label.htmlFor = el.id;
 	});
 };
 
--- a/docs/examples/add-another/script.js
+++ b/docs/examples/add-another/script.js
@@ -1,7 +1,0 @@
-if (typeof MOJFrontend.MultiFileUpload !== "undefined") {
-  new MOJFrontend.MultiFileUpload({
-    container: $(".moj-multi-file-upload"),
-    uploadUrl: "/ajax-upload",
-    deleteUrl: "/ajax-delete",
-  });
-}
--- a/package/moj/all.js
+++ b/package/moj/all.js
@@ -174,9 +174,13 @@ MOJFrontend.AddAnother.prototype.getNewItem = function() {
 
 MOJFrontend.AddAnother.prototype.updateAttributes = function(index, item) {
 	item.find('[data-name]').each(function(i, el) {
+    var originalId = el.id
+
 		el.name = $(el).attr('data-name').replace(/%index%/, index);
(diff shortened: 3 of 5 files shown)


package/moj/components/add-another/add-another.js
```diff
@@ -42,9 +42,13 @@ MOJFrontend.AddAnother.prototype.getNewItem = function() {
 
 MOJFrontend.AddAnother.prototype.updateAttributes = function(index, item) {
 	item.find('[data-name]').each(function(i, el) {
+    var originalId = el.id
+
 		el.name = $(el).attr('data-name').replace(/%index%/, index);
 		el.id = $(el).attr('data-id').replace(/%index%/, index);
-		($(el).siblings('label')[0] || $(el).parents('label')[0]).htmlFor = el.id;
+
+    var label = $(el).siblings('label')[0] || $(el).parents('label')[0] || item.find('[for="' + originalId + '"]')[0];
```
johnnolan, line 50: item.find isn't compatible with IE11, is this covered by polyfills elsewhere? https://caniuse.com/?search=find

gregtyler: Ah, it's not clear from the variable naming but `item` is a jQuery object, so it's [a different .find](https://api.jquery.com/find/)

johnnolan (approved):

PR #188, opened 2021-06-30
fix(assets): adds missing apple touch icon images

johnnolan:
Adds the apple touch icons for distribution in the package.

Sites that use this package seem to be adding these manually, this means they can be brought in from one place


gregtyler:
Might want to edit the commit message when merging: By being `feat`, I think it'll auto-bump a minor version (i.e. 0.3.0). If it were `fix`, we'd get a patch version bump (0.2.6) which is probably more appropriate.

johnnolan:
> Might want to edit the commit message when merging: By being `feat`, I think it'll auto-bump a minor version (i.e. 0.3.0). If it were `fix`, we'd get a patch version bump (0.2.6) which is probably more appropriate.

Have updated that now, good catch

gregtyler (approved):

PR #256, opened 2021-09-15
fix(add-another): shrink remove button

gregtyler:
The GOV.UK Design System extends buttons to 100% width on small screens. The Add another
component's "Remove" button is absolutely positioned, so it covers up content when it's at 100%
width.

This change shrinks the button back down to `width: auto`, so that it's only as wide as the text
and any content "underneath" can be seen.

I've also added right padding to the title so that it wraps before going underneath the remove
button.

This does reduce the clickable area for the button, which may make it harder for touchscreen users,
however there should not be any other clickable targets nearby, so users are unlikely to press the
wrong thing.

| Before | After |
|--------|-------|
| Screenshot 2021-09-15 at 10 59 21 | Screenshot 2021-09-15 at 10 59 52 |

### DIFF
--- a/src/moj/components/add-another/_add-another.scss
+++ b/src/moj/components/add-another/_add-another.scss
@@ -16,7 +16,7 @@
 
   &__title {
     float: left;
-    padding: 4px 0;
+    padding: 4px 100px 4px 0;
     width: 100%;
 
     & + .govuk-form-group {
@@ -28,6 +28,7 @@
     position: absolute;
     right: 0;
     top: 0;
+    width: auto;
   }
 
   &__add-button {


johnnolan:
Looks good but in my mind that button would drop below the text? However I gather there is someone who is better than me at UX who prefers this to the right?

gregtyler:
I don't disagree, and a focus order issue was raised in the DAC report. However, that would be a bigger change in terms of code and user impact, and I wanted to do something less risky to fix the current issue.

johnnolan:
Yeah that makes sense, thanks for the clarification and good point on impact to existing services

johnnolan (approved):

PR #261, opened 2021-09-21
Add case list pattern

kelliedesigner:
### Description of the change

- Add 'Case list' pattern, which communicates the pattern for making pieces of information about a case easier to scan, prioritise and action for users.
- Add link to the GitHub discussion for others to add to and iterate on the pattern

This pattern doesn't include working code, only screenshots and guidance. This pattern utilises existing GOV.UK components in describing how to use the pattern, reducing overhead and maintenance. 

|Case list pattern |
|---| 
|localhost_8080_patterns_case-list_(iPad Pro) (2) |

### Release notes

Users will be able to read and follow guidance on the 'Case list' pattern.

### DIFF
--- a/docs/_includes/layouts/patterns.njk
+++ b/docs/_includes/layouts/patterns.njk
@@ -49,6 +49,10 @@
               {
                 text: 'Task list pages',
                 href: ('/patterns/task-list' | url)
+              },
+              {
+                text: 'Case list pages',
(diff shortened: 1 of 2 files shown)


gregtyler (approved):
Love this. We've used a lot of these rules on services I've worked on, it's nice to get them down in writing 👍🏽

PR #281, opened 2021-11-11
Add a default aria-label for sub navigation component

tvararu:

Rather than opening an issue, I thought I'd use this PR for any necessary discussion, because the code change itself is small.

### Description of the change

[hmpps-manage-supervisions](https://github.com/ministryofjustice/hmpps-manage-supervisions) recently underwent a [DAC accessibility review](https://digitalaccessibilitycentre.org/). One of the issues they brought up in their audit was the lack of an `aria-label` on the sub navigation element in the app.

This PR sets a default `label` property on `mojSubNavigation` of "Secondary navigation region".

It looks like this in the accessibility inspector:

image

### Alternative designs

`label` could be made into a required property and produce a build error when it's not specified. This would break most apps that update to the latest version of the design system. I think silently filling in a sensible default is less friction for users of the design system.

### Possible drawbacks

An app updating to this version could conceivably already have a different navigation region which is coincidentally also called "Secondary navigation region", and a navigation region without a label, which would default to the same label. This wouldn't break anything, but it could slightly confuse assistive technology users, as there would now be two regions that are named the same. I think it is highly unlikely that someone would do this in practice, filling in a label for one region but not another.

### Verification process

I didn't spend much time on this yet as I'm waiting for feedback. Follow-up work on this PR:

- [x] Write automated tests if deemed necessary
- [x] Update documentation for this option

### Release notes

- Add a default aria-label for sub navigation component

### DIFF
--- a/docs/_includes/arguments/sub-navigation.md
+++ b/docs/_includes/arguments/sub-navigation.md
@@ -1,11 +1,11 @@
 ### Container
 
-| Name       | Type   | Required | Description                                                                  |
-| ---------- | ------ | -------- | ---------------------------------------------------------------------------- |
-| label      | string | No       | The `aria-label` to add to the `nav` container.                              |
-| items      | array  | Yes      | An array of navigation item objects. See [items](#items).                    |
-| classes    | string | No       | Classes to add to the `nav` container.                                       |
-| attributes | object | No       | HTML attributes (for example data attributes) to add to the `nav` container. |
+| Name       | Type   | Required | Description                                                                                |
+| ---------- | ------ | -------- | ------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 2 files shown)


gregtyler:
Hi Theodor, thanks for this. It sounds like a good suggestion to me, I'm very much in favour of sensible defaults, and I think that the possible drawback you mentioned is reasonable to accept on-balance.

In terms of follow-up work: there are no automated tests (yet), so there's nothing for you to worry about there. For documentation, I'd suggest updating the [arguments list](https://github.com/ministryofjustice/moj-frontend/blob/main/docs/_includes/arguments/sub-navigation.md) to include the label's new default value.

tvararu:
@gregtyler updated! Thanks for the quick feedback. :raised_hands:

gregtyler (approved):

michaelchambersjnr (approved):

PR #285, opened 2021-12-16
refactor(deps): make dependencies explicit

### LINKED ISSUE #147: Dependencies on GOV.UK are not clear
### Description

Some components have dependencies on functions, variables and mixins from `govuk-frontend`. For example, the cookie banner relies on [govuk-widh-container](https://github.com/ministryofjustice/moj-frontend/blob/master/src/moj/components/cookie-banner/_cookie-banner.scss#L19), [govuk-font](https://github.com/ministryofjustice/moj-frontend/blob/master/src/moj/components/cookie-banner/_cookie-banner.scss#L3), [govuk-colour](https://github.com/ministryofjustice/moj-frontend/blob/master/src/moj/components/cookie-banner/_cookie-banner.scss#L11) and more.

This is not an issue if all of the SCSS files from govuk-frontend are imported, but it is a problem if users only include the bits they need in their app. For example, a DWP service uses the following imports (which doesn't cover `govuk-width-container`):

```scss
@import "../../../../node_modules/govuk-frontend/govuk/base";
@import "../../../../node_modules/govuk-frontend/govuk/core/links";
@import "../../../../node_modules/govuk-frontend/govuk/components/button/index";
@import "../../../../node_modules/govuk-frontend/govuk/components/textarea/textarea";
@import "../../../../node_modules/@ministryofjustice/frontend/moj/all";
```

### Fixes

Two ways of fixing this spring to mind:
- Updating documentation to tell users what imports they need to make from govuk-frontend, either on a per-component basis or in the install instructions
- Importing any required files ourselves

The second option certainly seems the cleanest, but we'll need to explore it in depth to ensure there are no adverse side effects or cross-incompatibilities.

gregtyler:
`moj-frontend` has always depended on `govuk-frontend`. Not only does it use some of the classes in templates, but we explicitly use some of its variables and mixins. This PR makes that clear by making govuk-frontend an explicit dependency of the package and importing parts of its SCSS.

I removed the extension of `.govuk-textarea` in the Rich Text Editor component, with the extra class to be applied by CSS instead. This is to avoid outputting a definition for the govuk class in our bundled CSS file.

This PR also marks jQuery as a peer dependency, as it's necessary for the JavaScript to work but not bundled as part of the package.

This is a **breaking change** as the package now explicitly depends on govuk-frontend >=3. If you are using version 1.x.x or 2.x.x of govuk-frontend, you must upgrade that package before you can upgrade `moj-frontend`.

fixes #147


gregtyler:
I considered this as a breaking change, since it removes support for v2.x.x of `govuk-frontend`. However, we removed support in b88883413af85161c12ef042c51de245c0f35ae1, when we introduced colours that don't exist in the v2.

Instead, I've marked this as a `feat` since it will allow people to change their dependencies and build pipelines.

johnnolan (approved):
Nice

PR #294, opened 2022-01-06
feat(pagination): allow customisation of pagination results text

### LINKED ISSUE #293: Pagination: configurable "results" label
It would be useful to be able to specify the text used to display the count of results so that instead of always saying (for example) "Showing 1 to 50 of 90 results" it could instead say "Showing 1 to 50 of 90 documents".

https://github.com/ministryofjustice/moj-frontend/blob/58420e658979e143468cd4e4315900db98c69bc5/src/moj/components/pagination/template.njk#L32

gregtyler:
Allow users to use custom text in the pagination summary, defaulting to "results" if unspecified.

This allows slightly more customisation, with suitable fallback so that it's not a breaking change. The Pagination component is being [added to the GOV.UK Design System](https://github.com/alphagov/govuk-frontend/pull/2419) and GDS are taking a  broader look at customisation, design and accessibility. Once that component is generally available, we'll archive this one since it won't offer any additional functionality.

I also took this opportunity to update the Pagination component documentation, tidying up the example and documenting the Nunjucks macro options.

<details>
<summary>New documentation page with macro options</summary>

image
</details>

fixes #293


docs/_includes/arguments/pagination.md
```diff
@@ -0,0 +1,32 @@
...
+| href     | string  | No       | The link URL.                                            |
+| selected | boolean | No       | Set to `true` to indicate the currently selected page.   |
+
+### results
+
+| Name  | Type   | Required | Description                                          |
+| ----- | ------ | -------- | ---------------------------------------------------- |
+| count | number | Yes      | The total number of items.                           |
+| from  | number | Yes      | The number of the first item on the page.            |
+| count | number | Yes      | The number of the last item on the page.             |
```
liammcmurray, line 24: Is this right? The example uses 'to' for the number of the last item on the page.

liammcmurray (commented):
Looks good - just want to double-check that arg.

docs/_includes/arguments/pagination.md
gregtyler: Good catch Liam, I've fixed that now

liammcmurray (approved):
LGTM 🚀

PR #335, opened 2022-03-25
build: use node lts, sass and npm ci

gregtyler:
Build with the LTS version of node, allowing us to use the latest packages. Continue testing on all current LTSs and v10.

Replace node-sass with sass: node-sass is deprecated and the source of security vulnerabilities. Continue to use node-sass in tests to ensure builds work.

Use `npm ci` to get precise installs


.github/workflows/main.yml
```diff
@@ -34,7 +34,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 15.x]
+        node-version: [10.x, 12.x, 14.x, 16.x]
```
johnnolan, line 37: I have no doubt there is a reason, but out of interest, why are you building against all these versions instead of just 16?

johnnolan (approved):

.github/workflows/main.yml
gregtyler: It's because our main distribution is the raw sass files so we test that they compile on all LTS versions. Otherwise we could push an update that isn't compatible with node-sass on 12.x and break a downstream app

I'm not sure that's really a possibility, but it feels like a reasonable bit of due diligence

PR #344, opened 2022-04-10
fix attributes in pagination's template.njk

krisztin:
### Identify the bug

`id` and `aria-labelledby` attributes were swapped resulting in screen readers not reading out the pagination label

### Description of the change
Swapped the attributes

### Alternative designs

N/A

### Possible drawbacks
None

### Verification process

Tested with NVDA

### Release notes
N/A


gregtyler (approved):
Good spot, thanks!

PR #346, opened 2022-04-20
fix(pagination): improve accessibility of pagination component

### LINKED ISSUE #161: Fix Pagination labelling for screen readers
[The Pagination component](https://moj-design-system.herokuapp.com/components/pagination) provides direct links to pages which are simply numbers (i.e. "1", "2" and "3" in the MOJDS docs example). This is unhelpful to screen readers, as it provides little indication of what the link actually does.

We should add some screen reader only text (i.e. with class `govuk-visually-hidden`) to those links to indicate that they are page shortcuts:
```html
<li class="moj-pagination__item">
  <a class="moj-pagination__link" href="/page=1">
    <span class="govuk-visually-hidden">Page</span> 1
  </a>
</li>
```

We should also:
1. update the screen reader only text for labelling nav elements: we can remove the `<p>` and just use `<nav aria-label="Pagination navigation">` 
1. update the screen reader only text on prev/next links to "Previous [page]" rather than "Previous [set of pages]".
1. update the links in the MOJDS docs example to make more sense (i.e. all link to `?page=#`, including the prev/next links, to a page number that makes sense)

gregtyler:
Use aria-label to provide more descriptive label for pagination page links
    
Fix label content for previous/next links
    
Use aria-current to highlight that the current page is selected
    
Use ellipsis rather than three dots
    
Fixes #161

### DIFF
--- a/src/moj/components/pagination/template.njk
+++ b/src/moj/components/pagination/template.njk
@@ -1,29 +1,27 @@
-<nav class="moj-pagination {{- ' ' + params.classes if params.classes}}" aria-labelledby="pagination-label">
-
-  <p class="govuk-visually-hidden" id="pagination-label">Pagination navigation</p>
+<nav class="moj-pagination {{- ' ' + params.classes if params.classes}}" aria-label="Pagination navigation">
 
   <ul class="moj-pagination__list">
     {%- if params.previous %}
       <li class="moj-pagination__item  moj-pagination__item--prev">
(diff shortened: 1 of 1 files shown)


gregtyler:
Staging environment: https://moj-frontend-g6rhcr665-vercel-gregtylerco.vercel.app/

euhyung (approved):
fix to address DAC audit findings p53 re pagination component / adding support for SR

PR #348, opened 2022-04-20
fix(multi-file-upload): remove semantic HTML

gregtyler:
Use `<div>` instead of some lone `<dd>`s. This isn't semantic HTML so we should stop pretending it is.


gregtyler:
Staging environment: https://moj-frontend-8aqf7v6rs-vercel-gregtylerco.vercel.app/

euhyung (approved):
fix to address DAC audit issue p28 multi-file upload component / missing <dl> parent element. Decision that without values the files uploaded do not need to be contained by <dl>, therefore change to <div>, other acceptable solutions would be tables or unordered lists. ### Note: we should add a note in the documentation that when adding additional actions like a "delete button" of uploaded files there ought to be hidden text to distinguish all the delete buttons and reference them to the matching filename

PR #359, opened 2022-05-17
fix: markup Banner components as regions

gregtyler:
Banner components are given the region role and an appropriate label to identify them.

The label is taken from, in order or priority:
- The `iconFallbackText` label, if set
- The type of banner (e.g. "success", "warning", "information")
- The word "banner"

I also removed the visually-hidden insertion `iconFallbackText` to avoid duplication: if provided, this is now the region's label.

The only possible regression I can think of is if people have written `iconFallbackText` to flow as a full sentence with the text content, which would now be broken up. If someone strictly needed that behaviour, they can simply append the content to `html`.


euhyung:
as an aside, I think we might want to review all currentColour / svg implementations for the inverted contrast mode / windows
IMG_3237

euhyung (approved):

PR #360, opened 2022-05-17
fix: use inline SVGs

### LINKED ISSUE #162: SVGs don't match colour of parent text
Where we use inline SVGs (i.e. [Primary navigation with toggle search](https://moj-design-system.herokuapp.com/components/primary-navigation#toggle-search), [Rich text editor](https://moj-design-system.herokuapp.com/components/rich-text-editor) and [Timeline](https://moj-design-system.herokuapp.com/components/timeline)), the colour of the SVG icons is hard-coded into the SVG files themselves.

This means they don't adapt to the text colour around them changes, which makes the icons disappear in certain visual modes. For example, in Windows High Contrast mode, the background is changed to black and any black icons cannot be seen.

Instead, the SVGs should have the CSS attribute `fill: currentColor` on them so that they match the surrounding text colour. The individual SVGs will also need to be updated to remove the hard-coded fill colour.

gregtyler:
Rather than embedding SVGs as background images, use them inline. This allows them to keep their parents' fill colour, so they can always adapt to match the text.

When forced color mode is active (e.g. Windows High Contrast mode), we need to force the fill colour to match the context (i.e. text or link colour), otherwise it doesn't pass through properly on Windows.

This is an accessibility enhancement as it means that users who use tools like high-contrast mode to force changes in background and foreground colours will continue to be able to see iconography. Without it, icons are invisible if they're the same colour as the forced background colour.

Fixes #162

### DIFF
--- a/docs/examples/timeline/index.njk
+++ b/docs/examples/timeline/index.njk
@@ -19,25 +19,35 @@ arguments: timeline
     </span>
   </summary>
   <div class="govuk-details__text">
-    Ut quam nunc, vulputate ac metus pharetra, posuere maximus velit. <a href="https://www.google.com">Etiam nec interdum velit.</a> Suspendisse molestie lectus in eros ornare
+    Ut quam nunc, vulputate ac metus pharetra, posuere maximus velit. <a class="govuk-link" href="#">Etiam nec interdum velit.</a> Suspendisse molestie lectus in eros ornare
   </div>
 </details>
 {% endset -%}
 
 {%- set listHtml %}
   <ul class="govuk-list govuk-list--bullet">
-    <li><a class="govuk-link" href="#">Item 1</li>
-    <li><a class="govuk-link" href="#">Item 2</li>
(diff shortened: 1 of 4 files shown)


euhyung:
https://hidde.blog/making-single-color-svg-icons-work-in-dark-mode/ and https://adrianroselli.com/2018/03/svg-filtering-for-windows-high-contrast-mode.html - and svg needs to meet Success Criterium 1.1.1, ie either it is decorative (hide) or functional (wrap with accompanying link or standalone needs accessible text with it.

gregtyler:
Updated to include the icon in the link: I've had to float it to avoid some display issues with SVG+underline+whitespace.

I've also added rules for "forced color mode" which is the CSS terminology for High Contrast. This ensures that embedded SVGs are filled with the colour of their context (i.e. `windowText` or `linkText`) because fill colours don't cascade properly in Microsoft Edge. I've tested this in High Contrast mode on Windows using both Firefox and Microsoft Edge.

PR #362, opened 2022-05-26
docs: remove reference to TinyMCE

gregtyler:
It's not suitably accessible to specifically suggest. If we find an accessible rich text editor in the future, we'll add that in here instead.

### DIFF
--- a/docs/components/archived-components.md
+++ b/docs/components/archived-components.md
@@ -33,7 +33,7 @@ For more complex validation, use an accessible validation library.
 
 This component was deemed not sufficiently accessible to be used in live services and too complex to maintain.
 
-You should use an accessible rich text editor like [TinyMCE](https://www.tiny.cloud/tinymce/).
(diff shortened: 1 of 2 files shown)


euhyung:
got in touch with TinyMCE rich text editor parent company on 22.04.2022 and asked to see an audit or VPAT report / not available. They do have accessibility documentation and dedicated keyboard shortcuts, and an accessibility checker, but the later is part of the professional plan (paid). I've not been able to speak to an accessibility person at their end (asked to). Hence I vote to steer away from outright recommendation. 

Haven't reviewed CK editor, so again, not recommending product, but found that do have a VPAT and WCAG 2.0 (!) audit report (we'd be interested in report covering 2.1 Success Criteria, too) https://ckeditor.com/docs/ckeditor4/latest/guide/dev_wcag.html

euhyung (approved):

PR #366, opened 2022-06-24
docs: fix accessibility and usability issues

gregtyler:
Fixing the issues [identified in testing](https://docs.google.com/spreadsheets/d/1j7Ms7OcbMyEt4-AlZ-SGpUIl8XN3ylyFJixHQW9Z1XE/edit#gid=1536858132) (as much as possible)

Fixes:
- Adds a skip link before the site header
- Removes the link from "Ministry of Justice" (only "Pattern Library" should link to the homepage)
- Size of "Need help?" header now matches header level
- iframe examples now have `title` attribute set (rather than `aria-label`, which was intended to do the same but isn't the standard)
- "Back to top"  button scrolls with the page
- Crown copyright link is fixed
- Privacy policy removed as we don't have one
- Fix spelling of channel name on Patterns page

### DIFF
--- a/assets/sass/application.scss
+++ b/assets/sass/application.scss
@@ -16,6 +16,7 @@ $govuk-assets-path: "../" !default;
 @import "./components/back-to-top";
 @import "./components/copy-button";
 @import "./components/example";
+@import "./components/flex-container";
 @import "./components/markdown";
 @import "./components/masthead";
 @import "./components/panel";
--- a/assets/sass/components/flex-container.scss
+++ b/assets/sass/components/flex-container.scss
@@ -1,0 +1,4 @@
+/** Causes child elements to fill vertical space, allowing `position:sticky` to work as intended */
+.app-flex-container {
+  display: flex;
+}
--- a/assets/sass/components/flex-container.scss
+++ b/assets/sass/components/flex-container.scss
@@ -1,4 +1,6 @@
 /** Causes child elements to fill vertical space, allowing `position:sticky` to work as intended */
 .app-flex-container {
-  display: flex;
+  @include govuk-media-query($from: tablet) {
+    display: flex;
+  }
 }
--- a/docs/_includes/example.njk
+++ b/docs/_includes/example.njk
@@ -2,7 +2,7 @@
   <span class="app-example__new-window">
     <a href="{{ href | url }}" target="_blank">Open this example <span class="govuk-visually-hidden">({{ title }})</span> in a new window</a>
   </span>
-  <iframe class="app-example__frame" scrolling="auto" frameborder="0" height="{{ height }}" src="{{ href | url }}" aria-label="{{ title }}"></iframe>
+  <iframe class="app-example__frame" scrolling="auto" frameborder="0" height="{{ height }}" src="{{ href | url }}" title="{{ title }}"></iframe>
 </div>
 <div class="app-tabs" data-module="app-tabs">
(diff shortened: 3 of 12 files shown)


gregtyler:
🧪 Staging environment: https://moj-frontend-i0xxjgb5o-vercel-gregtylerco.vercel.app/

euhyung:
Hey, thank you very much - 4 comments 
- Removes the link from "Ministry of Justice" (only "Pattern Library" should link to the homepage) - initially I thought we'd just group this, think this would be clearer if it was "Ministry of Justice Pattern Library", but minor
- Size of "Need help?" header now matches header level - thanks, oversight of mine: in my list I've not mentioned that "Related resources" and the hidden "Support links" headings is marked up as heading level 2, but is the size of a heading level 3 - so we should change the code to match the level, ie heading level 3. There's one instance of missed heading mark up on "Upload files" page - first iframe example, "Upload receipt"
- "Back to top" button scrolls with the page - I think we need to reconsider the solution - if you start zooming into the page you'll see that on reflow content is dropping off page.
- Privacy policy removed as we don't have one - Can I have a link for the accessibility statement, please? Could be in the space of where Privacy policy link was previously.

euhyung:
thanks!! a few comments back 

 *  "Ministry of Justice Pattern Library" link - liking it
 * Size of "Need help?" header now matches header level - thanks, oversight of mine: in my list I've not mentioned that "Related resources" and the hidden "Support links" headings is marked up as heading level 2, but is the size of a heading level 3 - so we should change the code to match the level, ie heading level 3. There's one instance of missed heading mark up on "Upload files" page - first iframe example, "Upload receipt"
   * The element/font-sizes in the footers follow [the recommendations of the GDS component](https://design-system.service.gov.uk/components/footer/#footer-with-secondary-navigation) so I think we should take it up with them rather than drift from their standards
   * The iframe one is a little weird, but since it's previewing the example code it inevitably won't make sense in the context of the parent page. This is true of the GDS docs as well ([e.g. the various h1s on this page](https://design-system.service.gov.uk/components/checkboxes/))
       * From the above link I don't think the guidance is prescriptive about the link headings to be H2 - I think the heading levels have to be seen in context. I'm raising it because we still have a lot of H3s that visually seem exactly the same size as the H2 in the footer, which seems confusing. For example in "Prototyping", "Production", "Deploying ... " and a lot of Component and Pattern pages 
 *  "Back to top" button scrolls with the page - works! 
 *  Privacy policy removed as we don't have one - Can I have a link for the accessibility statement, please? Could be in the space of where Privacy policy link was previously.
     * I've added a link but the page just says "Coming soon" because I don't think we have the content yet. Is it better to link to a content-less page? Or not to add a link until we have the content sorted? 
       * Yeah, let's keep "Coming soon" for now, I've just noticed that GDS recommends to name the link itself "Accessibility" instead of "Accessibility statement" so let's amend that and we'll fill the page once we have content ready.

gregtyler:
I've updated the accessibility statement link/page title.

With regard to the footer headings, I switched to use the GDS template, which we should have been using all along and would have resolved the copyright link automatically for us. This has baked in the headings as h2s with `govuk-heading-m`, but I still don't think this is a problem. I agree that the Design System isn't prescriptive but every example uses that heading level/class combination, as does the template and [the GOV.UK website](https://www.gov.uk/). If it's an issue, we need to take it up with the GDS team so that they can change the template/develop guidance.

euhyung (approved):

PR #376, opened 2022-08-16
docs: fill out accessibility statement

gregtyler:


euhyung (changes requested):
Let's follow heading structure outlined in [sample statement ](https://www.gov.uk/government/publications/sample-accessibility-statement/sample-accessibility-statement-for-a-fictional-public-sector-website) that would include H1 Accessibility statement for MoJ Pattern Library and downsizing a few headings as to nest properly

euhyung (commented):
unnecessary line break on line 83 (users with a wide 
range of disabilities.) to be removed, please. 
Agree with your point about duplication. So suggest to keep all 3 bullet points under "Non-compliance and rephrase from line 61 onwards to: "We plan to address these issues by December 2022 apart from the unmarked iframe headings issue. This issue is a necessary effect of showing example snippets within pages and so cannot be fixed."

euhyung (approved):

PR #381, opened 2022-10-21
fix(scrollable): fix CSS shadow effect

gregtyler:
The `.scrollable-pane` class is supposed to add shadows to indicate that its contents are scrollable, but it is not suitably dynamic.

This CSS uses [scroll shadows](https://www.bram.us/2019/10/24/pure-css-scroll-shadows-vertical-horizontal/) to make the shadows dynamically appear on the left and right when there's overflow content there.

If there's no overflow content, or the container is scrolled all the way to one side, the shadows are hidden.

I removed the `.moj-scrollable-pane__wrapper` class because it's unnecessary.

### DIFF
--- a/docs/examples/patterns/filter-a-list/index.njk
+++ b/docs/examples/patterns/filter-a-list/index.njk
@@ -144,134 +144,129 @@ title: Filter a list pattern (example)
   </div>
 
   <div class="moj-scrollable-pane">
-
-    <div class="moj-scrollable-pane__wrapper">
-
-      {{ govukTable({
-        firstCellIsHeader: true,
-        head: [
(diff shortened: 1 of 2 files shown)


gregtyler:
This PR is just to fix the existing CSS. Next week I'll add documentation so that people can use it independently.

You can see it in action on [the preview environment](https://moj-frontend-oy0on19as-vercel-gregtylerco.vercel.app/patterns/filter-a-list): open the "Filter a list" pattern and click "Show Filter" and you'll see the shadows appear as you scroll left and right.

rwk-moj (approved):
LGTM! 👍

Issue #259, opened 2021-09-20
Sortable table is persistent


liammcmurray:
## Summary

Make the sorted data of a sortable table (https://design-patterns.service.justice.gov.uk/components/sortable-table/) stay sorted for the duration of a user's session.

## Motivation

Users interacting with a sorted table refer back to it as part of task completion, and expectation is that it remains in the state they left it in.

## Describe alternatives you've considered

Users can re-sort, but this is not a great user experience. It would be much better for work done not to be 'undone' in this instance.

## Additional context

The persistent pattern is particularly necessary where the sortable table contains links, as users get 'lost' when returning to the screen.

gregtyler:
We could add a persistent option to sortable (so that it's not default behaviour, for backwards compatibility) that:

- Stores the column and sort order whenever the table is sorted
- Uses the stored column and order when the page is loaded

We'd need the user to pass in an ID to store it under so that if there are multiple Sortables in an application they don't all persist with the same setting (but by sharing an ID, you could link two identical-in-practice tables).

PR #406, opened 2022-11-09
fix(sortable-table): Sort the columns of the sortable table on page load

smcveigh941:
When using the sortable table, the attribute `"aria-sort": "ascending"` can be set in the nunjucks document. This makes the page load with the column header marked as ascending or descending depending on what the consumer has set. 

However, the values in the column are not sorted initially. This PR sorts the columns on the initial load of the page using the sort direction which is defined in the nunjucks file.

### DIFF
--- a/src/moj/components/sortable-table/sortable-table.js
+++ b/src/moj/components/sortable-table/sortable-table.js
@@ -11,6 +11,7 @@ MOJFrontend.SortableTable = function(params) {
 	this.body = this.table.find('tbody');
 	this.createHeadingButtons();
 	this.createStatusBox();
+  this.initialiseSortedColumn();
 	this.table.on('click', 'th button', $.proxy(this, 'onSortButtonClick'));
 };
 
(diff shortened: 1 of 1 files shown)


gregtyler:
Thanks @smcveigh941, this is really helpful. It'll also be useful in the feature we're considering to make the sorting (optionally) persistent (#259) which is nice.

A few things I considered in review:
- Accessibility: the sorting happens immediately so shouldn't need an announcement. This also ensures that the sorting _actually matches the ARIA attribute_ which is a definite improvement.
- User experience: if the table indicates (both visually and through ARIA attributes) that it's sorted, users would expect it to be sorted.
- Testing: we don't have component-level testing yet, so no criteria to add any tests.
- Unexpected errors: since we add the `data-index` attribute and the code specifically checks for `aria-sort`, there shouldn't be any nasty surprises.
- Code style: we don't apply any formal code style rules, but this generally follows the patterns used in the rest of the library

gregtyler (approved):

PR #412, opened 2022-11-22
fix(filter-toggle-button): make `bigModeMediaQuery` optional

gregtyler:
Currently if `bigModeMediaQuery` isn't set then the component will always initialise in "small" mode, where the filter bar is hidden, even if `startHidden` is set to false or not specified.

This change makes `bigModeMediaQuery` optional so that, if it isn't set, `startHidden` is adhered to.

`bigModeMediaQuery` still takes priority over `startHidden` if both are set.

BREAKING CHANGE: If you specify neither `bigModeMediaQuery` nor `startHidden`, the filter will now be initially open on larger screens. To avoid this behaviour, add the `startHidden: true` property.


gregtyler:
🚀 Deploying to [staging environment](https://moj-frontend-staging.apps.live.cloud-platform.service.justice.gov.uk/)

Issue #519, opened 2023-10-12
Review MOJ Pattern Library against WCAG 2.2 criteria


gregtyler:
## Summary

Now that WCAG 2.2 is published, we need to review the components and patterns of the Pattern Library—and the documentation site itself—against the new criteria. We should then fix any issues we find, or offer additional guidance if we identify implementation risks.

## New criteria in WCAG 2.2

Based on [w3.org](https://www.w3.org/WAI/standards-guidelines/wcag/new-in-22/)

- 2.4.11 Focus Not Obscured (Minimum) (AA)
- 2.4.12 Focus Not Obscured (Enhanced) (AAA)
- 2.4.13 Focus Appearance (AAA)
- 2.5.7 Dragging Movements (AA)
- 2.5.8 Target Size (Minimum) (AA)
- 3.2.6 Consistent Help (A)
- 3.3.7 Redundant Entry (A)
- 3.3.8 Accessible Authentication (Minimum) (AA)
- 3.3.9 Accessible Authentication (Enhanced) (AAA)

gregtyler:
[in progress...]

## Components

### 2.4.11 and 2.4.12 Focus Not Obscured

- The [Primary navigation search toggle](https://design-patterns.service.justice.gov.uk/components/primary-navigation/#toggle-search) may do this if there is interactive content below the pop-out search box
  - I think the most consistent solution to this is to minimise the pop-out when it loses focus
  - Otherwise, we could tell implementers to not position content there, or to minimise the box if content is focused. But that would push a lot of work onto implementers, and is likely to not work consistently.

| Layout | With pop-out expanded (obscured link has focus) |
| ---- | ---- |
|  ![image](https://github.com/ministryofjustice/moj-frontend/assets/100852/5f35e4a9-ae90-4995-a092-07270a771145) |  ![image](https://github.com/ministryofjustice/moj-frontend/assets/100852/e2e9e1d3-4f00-42f8-80e6-e390abc800c4) |

- The [Button menu](https://design-patterns.service.justice.gov.uk/components/button-menu/#left-aligned-grey-toggle-menu) also has a pop-out but it minimises when the focus is lost, ensuring that focused content is not obscured
- The [Scrollable pane](https://design-patterns.service.justice.gov.uk/components/scrollable-pane/) supports off-screen content but the browser will bring it into view if focussed

### 2.4.13 Focus Appearance

[...]

PR #520, opened 2023-10-18
feat(primary-navigation): auto-hide search toggle when it loses focus

gregtyler:
WCAG's [Focus Not Obscured (Minimum) success criterion](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum) states:
> When a user interface component receives keyboard focus, the component is not entirely hidden due to author-created content.

The Primary navigation search toggle fails this criterion as it continues to overlay interactive content when it loses focus. This can lead to focused content being entirely hidden.

This change auto-closes the toggle when it loses focus, either through another element being focused, or through the user clicking elsewhere on the page. It can be manually re-opened by pressing the toggle button again.

[Interactive example of old and new behaviour](https://mojpl-520-autoclose.glitch.me/)

See #519 for initial conversation on this topic

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -41,7 +41,7 @@ module.exports = function (eleventyConfig) {
   );
 
   eleventyConfig.addShortcode("example", function (exampleHref, height) {
-    const { data, content: nunjucksCode } = matter(
+    let { data, content: nunjucksCode } = matter(
       fs
         .readFileSync(
           path.join(__dirname, "docs", exampleHref, "index.njk"),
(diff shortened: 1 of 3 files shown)


euhyung:
One additional accessibility comment re labelling: At the moment there's a hidden "Search" label used for the Search form field. As it's not labelled by "Find Case" (in this case a button), wondering if there's a case of using the hint text for this instance to make that label more instructive. Aware that hint text is not usually the go-to label for the form field. Wondering if anyone has found issues when using this component re labelling? Are there any teams aware of labelling issues for the toggle search?

gregtyler:
Done in #565

PR #528, opened 2023-10-27
feat: add support for v5 of the Design System

gregtyler:
**This is a work-in-progress to allow services using the MOJPL to test v5 of the Design System**

Allows the Pattern Library to be used with v5 of the Design System, and uses v5 by default in the documentation.

## Changes

We currently continue to support IE8 by providing a copy of the `govuk-if-ie8` mixin which was removed in v5 of the Design System. This is so we can continue to support services which still support IE8 and use v5 of the Design System, though we may also drop support in a subsequent update.

## Status

The build is currently failing because of a private peer dependency in govuk-frontend. That is being moved to a dev dependency, which should unblock this build.

### DIFF
--- a/docs/_includes/layouts/base.njk
+++ b/docs/_includes/layouts/base.njk
@@ -15,8 +15,7 @@
     <meta name="description" content="Design your service using MoJ styles, components and patterns">
     <meta property="og:image" content="{{ '/assets/images/moj-opengraph-image.png' | url }}">
 
-    <!--[if lte IE 8]><link href="{{ '/assets/stylesheets/application-ie8.css' | url }}" rel="stylesheet"><![endif]-->
-    <!--[if gt IE 8]><!--><link href="{{ '/assets/stylesheets/application.css' | url }}" rel="stylesheet"><!--<![endif]-->
+    <link href="{{ '/assets/stylesheets/application.css' | url }}" rel="stylesheet">
     <title>{% block title %}{{ title }} - MoJ Pattern Library{% endblock %}</title>
 
     <script type="module" src="https://www.googletagmanager.com/gtag/js?id=G-VTGX4YLSVL" async></script>
--- a/docs/_includes/layouts/base.njk
+++ b/docs/_includes/layouts/base.njk
@@ -16,7 +16,7 @@
     <meta property="og:image" content="{{ '/assets/images/moj-opengraph-image.png' | url }}">
 
     <link href="{{ '/assets/stylesheets/application.css' | url }}" rel="stylesheet">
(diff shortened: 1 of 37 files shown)


gregtyler:
Because `govuk-frontend` has moved their files (into a `dist` folder), our current code cannot support both ≤v4 and ≥v5. To mitigate this I've removed the direct import. This means that anyone building our SCSS into their bundle needs to include the govuk-frontend code beforehand which I believe should be commonplace already and, if not, is a one-line fix.

This is **a breaking change** and makes this a major release. That means this also might be a good opportunity to remove support for IE8 in line with `govuk-frontend` (our IE8 support is already weakened by some govuk functions removing support themselves).

ttbarnes:
@gregtyler Anything I can do to help progress this?

gregtyler:
@ttbarnes If you can try out [the pre-release](https://github.com/ministryofjustice/moj-frontend/releases/v2.0.0-beta.1) on your service to check it works alongside v5 of the GOV.UK Design System, any feedback would be helpful. I haven't seen much feedback—positive or negative—but if we don't hear anything negative soon then hoping to release in the first weeks of 2024.

ttbarnes:
Thank you @gregtyler great stuff. I have just tested the pre-release - we're only using a couple of MOJ components, but all seems great including JS/SCSS compilation alongside the latest GOV design system 👍

thomasleese:
We're using the timeline component in our service and it also looks good under this pre-release.

docs/_includes/layouts/base.njk
```diff
@@ -15,16 +15,15 @@
     <meta name="description" content="Design your service using MoJ styles, components and patterns">
     <meta property="og:image" content="{{ '/assets/images/moj-opengraph-image.png' | url }}">
 
-    <!--[if lte IE 8]><link href="{{ '/assets/stylesheets/application-ie8.css' | url }}" rel="stylesheet"><![endif]-->
-    <!--[if gt IE 8]><!--><link href="{{ '/assets/stylesheets/application.css' | url }}" rel="stylesheet"><!--<![endif]-->
-    <title>{% block title %}{{ title }} - MoJ Design System{% endblock %}</title>
+    <link href="{{ '/assets/stylesheets/application.css' | url }}" rel="stylesheet">
+    <title>{% block title %}{{ title }} - MoJ Pattern Library{% endblock %}</title>
```
robertjmccarthy, line 19: Change 'MoJ Pattern Library' to 'MoJ Design System'.

robertjmccarthy (approved):

PR #545, opened 2023-12-15
feat(navigation): enhance hover styles of navigation links

### LINKED ISSUE #447: Colour contrast for Primary navigation hover is not AA compliant
### Description

When hovering over a navigation item on the [Primary navigation component](https://design-patterns.service.justice.gov.uk/components/primary-navigation/), the link text is `#5694ca` on a background of `#f3f2f1`. At a colour contrast ratio of 2.89:1, this is not AA accessible.

image

### Versions

Discovered in v1.6.4, but presumably has been a problem for much longer.

### Additional Information

The [GOV.UK Design System documentation](https://design-system.service.gov.uk/) (which I think the component was originally forked from) is now using a darker colour and additional underline on hover, which fixes the contrast issue and also ensures users who aren't able to detect the contrast difference can still see a hover effect.

image

gregtyler:
Applies to Primary navigation, Sub navigation and Side navigation.

Use the standard GDS link hover colour to ensure links colour contrast is sufficient for
readability.

Ensure borders follow suit, and ensure they're always on the same side of the element.

Fixes #447

### DIFF
--- a/src/moj/components/primary-navigation/_primary-navigation.scss
+++ b/src/moj/components/primary-navigation/_primary-navigation.scss
@@ -62,7 +62,7 @@
   }
 
   &:hover {
-    color: govuk-tint($govuk-link-colour, 25);
+    color: $govuk-link-hover-colour;
   }
 
   &:focus {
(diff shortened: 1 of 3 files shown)


robertjmccarthy (approved):

gregtyler:
Done in #565

PR #550, opened 2024-01-03
feat(date-picker): add date picker component

gregtyler:
Can be initialised in Nunjucks with the `mojDatePicker` macro, or in JavaScript with the `data-module="moj-date-picker"` attribute.

Fixes MDS-72

### DIFF
--- a/src/moj/components/date-picker/_date-picker.scss
+++ b/src/moj/components/date-picker/_date-picker.scss
@@ -1,0 +1,278 @@
+.moj-datepicker {
+  position: relative;
+
+  &--fixed-width {
+    .moj-datepicker-input__wrapper {
+      width: 215px;
+    }
+  }
+
+  &__dialog {
+    box-shadow: 1px 1px 4px rgba(0, 0, 0, 0.15);
+
+    background-color: govuk-colour('white');
+    clear: both;
+    display: none;
+    padding: 8px;
+    outline: 1px solid $govuk-border-colour;
+    outline-offset: -1px;
+    position: static;
+    top: 0;
+    transition: background-color 0.2s, outline-color 0.2s;
+    width: 280px;
+    z-index: 2;
+
+    &__header {
+      position: relative;
+      text-align: center;
+      margin-bottom: 5px;
+
+      > :nth-child(1) {
+        position: absolute;
+        left: 5px;
+        top: -2px;
+
+        > :nth-child(2) {
+          margin-left: 4px;
+        }
+      }
+
+      > :nth-child(3) {
+        position: absolute;
+        right: 5px;
+        top: -2px;
+
+        > :nth-child(1) {
+          margin-right: 4px;
+        }
+      }
+    }
+
+    &__title {
+      font-size: 16px;
+      padding: 8px 0;
+      margin: 0 !important;
+    }
+
+    &__navbuttons {
+      button {
+        background-color: transparent;
+        color: $govuk-text-colour !important;
+        min-height: 40px;
+        margin: 0;
+        padding: 4px 4px 0 4px;
+        min-width:  32px;
+        border: none;
+        display: inline-block;
+        cursor: pointer;
+        outline: none;
+
+        .moj-datepicker-icon {
+          height: 32px;
+          padding: 0;
+          position: static;
+          width: 24px;
+        }
+
+        &:hover {
+          background-color: rgba(govuk-colour('yellow'), .5);
+        }
+
+        &:focus {
+          background-color: $govuk-focus-colour;
+          border-bottom: 4px solid govuk-colour('black');
+        }
+      }
+    }
+
+    &__table {
+      border-collapse: collapse;
+
+      tbody:focus-within {
+        outline: 2px solid $govuk-focus-colour;
+      }
+
+      td {
+        border: 0;
+        margin: 0;
+        outline: 0;
+        padding: 0;
+      }
+
+      th {
+        font-size: 16px;
+        color: $govuk-text-colour;
+      }
+
+      button {
+        background-color: transparent;
+        border-width: 0;
+        color: $govuk-text-colour;
+        min-height: 40px;
+        margin: 0;
+        padding: 0;
+        min-width: 40px;
+
+        font-size: 16px;
+
+        &:hover {
+          outline: 3px solid rgba(0,0,0,0);
+          color: $govuk-text-colour;
+          background-color: rgba(govuk-colour('yellow'), .5);
+          box-shadow: none;
+          text-decoration: none;
+          -webkit-box-decoration-break: clone;
+          box-decoration-break: clone;
+          cursor: pointer;
+        }
+
+        &:focus {
+          outline: 3px solid rgba(0,0,0,0);
+          color: $govuk-focus-text-colour;
+          background-color: $govuk-focus-colour;
+          border-bottom: 4px solid govuk-colour('black');
+          padding-top: 4px;
+          text-decoration: none;
+          -webkit-box-decoration-break: clone;
+          box-decoration-break: clone;
+        }
+
+        &[disabled="true"] {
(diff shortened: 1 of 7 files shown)


src/moj/components/date-picker/_date-picker.scss
```diff
@@ -0,0 +1,278 @@
+.moj-datepicker {
+  position: relative;
+
+  &--fixed-width {
+    .moj-datepicker-input__wrapper {
+      width: 215px;
+    }
+  }
+
+  &__dialog {
```
gregtyler, line 10: We should use `@include govuk-font(16);` on this to make sure it has the right font face and size

src/moj/components/date-picker/_date-picker.scss
```diff
@@ -0,0 +1,278 @@
+.moj-datepicker {
+  position: relative;
+
+  &--fixed-width {
+    .moj-datepicker-input__wrapper {
+      width: 215px;
+    }
+  }
```
gregtyler, line 8: Rather than fixing the width of the wrapper, we should let people set the input width with the existing GDS classes (e.g. `govuk-input--width-10`)

src/moj/components/date-picker/date-picker.js
```diff
@@ -0,0 +1,596 @@
...
+  // populates calendar with initial dates, avoids Wave errors about null buttons
+  this.updateCalendar()
+}
+
+Datepicker.prototype.createDialogMarkup = function (titleId) {
+  return `<div class="moj-datepicker__dialog__header ">
+        <div class="moj-datepicker__dialog__navbuttons">
+            <button class="js-datepicker-prev-year" aria-label="previous year" data-button="button-datepicker-prevyear">
+                <span class="govuk-visually-hidden">Previous year</span>
+                <svg focusable="false" class="moj-datepicker-icon" aria-hidden="true" role="img"><use href="/assets/images/icons.stack.svg#double_chevron_left"></use></svg>
```
gregtyler, line 163: We don't have `icons.stack.svg` so will need to define these SVGs inline

src/moj/components/date-picker/date-picker.js
```diff
@@ -0,0 +1,596 @@
...
+  document.body.addEventListener('mouseup', event => this.backgroundClick(event))
+
+  // populates calendar with initial dates, avoids Wave errors about null buttons
+  this.updateCalendar()
+}
+
+Datepicker.prototype.createDialogMarkup = function (titleId) {
+  return `<div class="moj-datepicker__dialog__header ">
+        <div class="moj-datepicker__dialog__navbuttons">
+            <button class="js-datepicker-prev-year" aria-label="previous year" data-button="button-datepicker-prevyear">
```
gregtyler, line 161: `js-datepicker-prev-year` needs to be `moj-` namespaced

src/moj/components/date-picker/date-picker.js
```diff
@@ -0,0 +1,596 @@
...
+  }
+}
+
+Datepicker.prototype.openDialog = function () {
+  // display the dialog
+  this.dialogElement.style.display = 'block'
+  this.dialogElement.classList.add('moj-datepicker__dialog--open')
+
+  // position the dialog
+  this.dialogElement.style.left = `${this.$input.offsetWidth + 16}px`
```
gregtyler, line 400: If the button is on the right-hand side of the window, this overflows and stretches the page width (or is inaccessible if in an unscrollable container)

src/moj/components/date-picker/template.njk
```diff
@@ -0,0 +1,64 @@
...
+  {% endif %}
+  <div class="moj-datepicker-input__wrapper">
+    {{ govukInput({
+      classes: "govuk-input moj-js-datepicker-input",
+      id: params.id,
+      name: params.name,
+      value: params.value,
+      autocomplete: "off",
+      attributes: {
+        "maxlength": "10",
```
gregtyler, line 46: GDS advise against creating hard-restrictions for users like this, it should be handled by post-submission validation instead

src/moj/components/date-picker/date-picker.js
```diff
@@ -0,0 +1,596 @@
...
+  this.picker.selectDate(this.date)
+
+  event.stopPropagation()
+  event.preventDefault()
+}
+
+DSCalendarDay.prototype.keyPress = function (event) {
+  let calendarNavKey = true
+
+  switch (event.keyCode) {
```
gregtyler, line 555: `keyCode` is deprecated, but the replacement [code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#browser_compatibility) doesn't have sufficient support (most notably, Chrome for Android doesn't support it)

src/moj/components/date-picker/date-picker.js
```diff
@@ -0,0 +1,596 @@
...
+    this.closeDialog(event)
+  })
+  this.okButton.addEventListener('click', () => this.selectDate(this.currentDate))
+
+  const dialogButtons = this.dialogElement.querySelectorAll('button:not([disabled="true"])')
+  // eslint-disable-next-line prefer-destructuring
+  this.firstButtonInDialog = dialogButtons[0]
+  this.lastButtonInDialog = dialogButtons[dialogButtons.length - 1]
+  this.firstButtonInDialog.addEventListener('keydown', event => this.firstButtonKeyup(event))
+  this.lastButtonInDialog.addEventListener('keydown', event => this.lastButtonKeyup(event))
```
gregtyler, line 148: nit, but these methods shoul be renamed `Keydown`

src/moj/components/date-picker/date-picker.js
```diff
@@ -0,0 +1,596 @@
...
+                <svg focusable="false" class="moj-datepicker-icon" aria-hidden="true" role="img"><use href="/assets/images/icons.stack.svg#double_chevron_right"></use></svg>
+            </button>
+        </div>
+      </div>
+
+      <table class="moj-datepicker__dialog__table  js-datepicker-grid" role="grid" aria-labelledby="${titleId}">
+      <caption class="moj-datepicker__dialog__table-caption">You can use the arrow keys to select a date</caption>
+      <thead>
+          <tr>
+          <th scope="col" abbr="Monday">Mo</th>
```
gregtyler, line 191: Should probably add `aria-label` to these to avoid screenreaders announcing "Mo"

src/moj/components/date-picker/date-picker.js
```diff
@@ -0,0 +1,596 @@
...
+  let dayCount = 0
+  for (let i = 0; i < 6; i++) {
+    // create row
+    const row = tbody.insertRow(i)
+
+    for (let j = 0; j < 7; j++) {
+      // create cell (day)
+      const cell = document.createElement('td')
+      const dateButton = document.createElement('button')
+      dateButton.dataset.form = 'date-select'
```
gregtyler, line 113: I'm not sure what `.dataset.form` is doing here or if it's needed.

We might need to add `aria-label` to explain what the button does (e.g. "28th January 2024"), but I'm not sure what state of the art is here

gregtyler (commented):
Some general thoughts:
- The popup doesn't close on clickaway, which could lead to a violation of WCAG 2.2 Focus Not Obscured
- If JavaScript isn't available, it doesn't gracefully degrade as the button is still visible. The button should be added by JavaScript rather than in the template HTML.
  - In general, I think the template shouldn't contain form groups, error messages, etc. It should just be `{{ govukInput({attributes: {'data-module': 'moj-date-picker'}}) }}`
- Whilst we haven't back-filled it to existing components, we should make sure this has localisation support for all text on release

src/moj/components/date-picker/date-picker.js
chrispymm: Based on the link, support looks pretty good across the board.  Chrome for Android looks to have had support from 2016.  Maybe the page was incorrect and has since been updated.

PR #557, opened 2024-01-15
feat: allow Elements to be used in components' config

gregtyler:
Rather than requiring that all elements used as components config are jQuery objects, allow `Element` objects to be passed too.

This means that you can use the bundled version of moj-frontend without installing jQuery yourself.


gregtyler:
#565

PR #887, opened 2024-10-29
fix(banner): fix issue with classes not working if banner type provided

### LINKED ISSUE #886: Banner classes parameter does not work when type parameter is set
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of the following:
  - Reproduced the problem
  - Followed all applicable steps in the debugging guide
  - Checked the FAQs on the message board for common solutions
  - Checked that your issue doesn't exist: https://github.com/ministryofjustice/moj-frontend/issues
  - Checked that there is not already a package that provides the described functionality

### Description

Banner classes do not work if `param.type` is configured. I assume this is a bug

### Steps to Reproduce

1. Create a banner component
2. Add a type: 
```
{{ mojBanner({
  type: 'information',
  html: bannerHtml
}) }}
```
3. Check type is rendering correctly
3. Add a class
```
{{ mojBanner({
  type: 'information',
  html: bannerHtml
  classes:'aClassHere'
}) }}
```
4. Note that the class is not applying correctly
5. Remove the type
```
{{ mojBanner({
  html: bannerHtml
  classes:'aClassHere'
}) }}
```
6. Note the class is now rendering correctly

**Expected behaviour:** [What you expect to happen]
Setting the `classes` attribute should be independent of the `type` attribute behaviour.

**Actual behaviour:** [What happens]
Setting the `type` attribute means `classes` can no longer be set

**Reproduces how often:** [What percentage of the time does it reproduce?]
Every time

### Versions
Latest (3.0.0) -- Present for at least 2 years

### Additional Information

 - Existing implementation appears to have an `else` and `endIf` block in a place that disables `classes` if `type` is set:
```
// Existing implementation (spacing added for readability)
<div class=
  "moj-banner
  {% if params.type == 'success' %} moj-banner--success
  {% elif params.type == 'warning' %} moj-banner--warning
  {% else %}           // This else and endIf below cause the bug
  {{- ' ' + params.classes if params.classes}}
  {% endif %}"
  ...
>
```
- I've raised a PR that includes some refactoring -- The first commit is a simple fix, the latter commit provides a refactor in line with recent `button-menu`.

- PR for fix [here](https://github.com/ministryofjustice/moj-frontend/pull/887)

AlexBramhill:
### Identify the bug

[Link to the issue describing the bug that you're fixing.
](https://github.com/ministryofjustice/moj-frontend/issues/886)

### Description of the change

 - Existing implementation appears to have an `else` and `endIf` block in a place that disables `classes` if `type` is set:
```
// Existing implementation (spacing added for readability)
<div class=
  "moj-banner
  {% if params.type == 'success' %} moj-banner--success
  {% elif params.type == 'warning' %} moj-banner--warning
  {% else %}           // This else and endIf below cause the bug
  {{- ' ' + params.classes if params.classes}}
  {% endif %}"
  ...
>
```
- Updated and refactored file to address this bug, and increase readability following pattern of declaring classes logic above component as per recent `button-menu` commit to main 

### Alternative designs

- I have a commit on this branch (hash `5742f092b7cc670d6ab453459f03c1252964d07e`) that does not do the refactor, but addresses the bug

### Possible drawbacks

If this is an intended business logic, then this change is not needed.

### Verification process

Manual testing:
- [x] check can set `type` and component reflects this
- [x] check can set `classes` and component reflects this  
- [x] check can set `type` and `classes`  and component reflects this
- [x] check if `type` and `classes` contradict, `type` takes precedence

### Release notes

Fixed #886 where if a banner type is present, banner classes could not be set


src/moj/components/banner/template.njk
```diff
@@ -13,4 +35,4 @@
     {{- params.html | safe if params.html else params.text -}}
   </div>
 
-</div>
+</div>
```
AlexBramhill, line 38:
```suggestion
</div>

```

src/moj/components/banner/template.njk
```diff
@@ -1,4 +1,26 @@
...
+
+{%- if params.type == 'success' %}
+  {%- set classNames = classNames.concat("moj-banner--success") %}
+{%- elif params.type == 'warning' %}
+  {%- set classNames = classNames.concat("moj-banner--warning") %}
+{%- elif params.type == 'information' %}
+  {%- set classNames = classNames.concat("moj-banner--information") %}
+{%- endif %}
+
+{%- classNames.concat(params.classes) if params.classes %}
```
chrispymm, line 12:
```suggestion
{%- set classNames = classNames.concat(params.classes) if params.classes %}
```

chrispymm (changes requested):
Thanks for the PR.  I'm pretty sure this is a bug and not intended behaviour.

Changes look good and the refactor is nice too.

Spotted one issue, where I think you're missing  `set`.  If you can fix that up, then I can get this merged.

src/moj/components/banner/template.njk
AlexBramhill: Updated, thanks :)

src/moj/components/banner/template.njk
```diff
@@ -1,4 +1,26 @@
...
+  {%- set classNames = classNames.concat("moj-banner--success") %}
+{%- elif params.type == 'warning' %}
+  {%- set classNames = classNames.concat("moj-banner--warning") %}
+{%- elif params.type == 'information' %}
+  {%- set classNames = classNames.concat("moj-banner--information") %}
+{%- endif %}
+
+{%- set classNames = classNames.concat(params.classes) if params.classes %}
+
+{%- set classNames = classNames.join(' ') -%}
```
AlexBramhill, line 14: This is causing the build to fail -- I'm looking into this, but might revert the refactor if I cannot find an immediate soluion (to fix the bug as a priority)

src/moj/components/banner/template.njk
```diff
@@ -1,4 +1,26 @@
...
+
+{%- if params.type == 'success' %}
+  {%- set classNames = classNames.concat("moj-banner--success") %}
+{%- elif params.type == 'warning' %}
+  {%- set classNames = classNames.concat("moj-banner--warning") %}
+{%- elif params.type == 'information' %}
+  {%- set classNames = classNames.concat("moj-banner--information") %}
+{%- endif %}
+
+{%- set classNames = classNames.concat(params.classes) if params.classes %}
```
chrispymm, line 12:
```suggestion
{%- if params.classes %}
{%- set classNames = classNames.concat(params.classes)  %}
{%- endif %}
```
This solves the issue.  I think if `params.classes` was empty it was still running the `set`. Wrapping it in the if, avoids this.

chrispymm (changes requested):

src/moj/components/banner/template.njk
AlexBramhill: Thanks -- I had just figured that out too! Seems to be working my end from what I can see :) Thank you for your review and comments Chris

chrispymm: No worries.  Thanks for the PR.

If you create another PR in the future, the actions will run automatically so you'll see any issues sooner.  It's just because you're a first time contributor that they require permission to run.

chrispymm (approved):

PR #928, opened 2024-11-15
fix(scss): amend sass loadpaths to include node_modules directory

### LINKED ISSUE #927: Design system sass import broken in Rails apps
`all.sccs` in our package contains an import with the path `@import node_modules/govuk-frontend/dist/govuk/base` 

This path seems to be causing some users issues when including it into their projects, and they cannot build the design system css.

This issues seems to primarily affect rails projects: 

https://github.com/rails/cssbundling-rails/issues/101
https://github.com/rails/cssbundling-rails/commit/9075819d32c7e1103b1ee7c237e79001547066af

cssbundling rails includes node_modules in the load path automatically, so having it in the import path in the sass file causes it to fail.

The recommendation from govuk, is to use the full path for imports (https://frontend.design-system.service.gov.uk/import-css/#import-specific-parts-using-sass)

Fixing this for rails projects, could cause issues for others...

chrispymm:
This allows us to remove the node modules prefix from the govuk scss import in all.scss. Which would resolve loading issues for users.

BREAKING CHANGE: This PR changes the loadpath for govuk frontend.  You will now need to ensure your project SASS configuration includes node_modules within its loadpaths in order to compile, however this is often default behaviour in many build tools.

fix #927

### DIFF
--- a/gulp/dist.js
+++ b/gulp/dist.js
@@ -27,7 +27,9 @@ gulp.task("dist:javascript", () => {
 gulp.task("dist:css", () => {
   return gulp
     .src("gulp/dist-scss/*.scss")
-    .pipe(sass())
+    .pipe(sass({
+      includePaths: ["node_modules"]
+    }))
     .pipe(postcss([autoprefixer, cssnano]))
     .pipe(
       rename((path) => ({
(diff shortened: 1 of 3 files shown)


starswan:
Hi, there is one module that also use node_modules - cookie_banner. It would be nice to fix that one at the same time.

Issue #914, opened 2024-11-11
Mutli-select component doesn't allow multiples on same page


starswan:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of the following:
  - Reproduced the problem
  - Followed all applicable steps in the debugging guide
  - Checked the FAQs on the message board for common solutions
  - Checked that your issue doesn't exist: https://github.com/ministryofjustice/moj-frontend/issues
  - Checked that there is not already a package that provides the described functionality

### Description

Put multiple multi-select components on the same page at once

### Steps to Reproduce

Attempt to interact with second multi-select 'select all' button to either select or deselect all items in second component 

**Expected behaviour:** 

Second multi-select selects and deselects all items

**Actual behaviour:** [What happens]

It appears to interact with the first list instead

**Reproduces how often:** 

100%

### Versions

5.71

### Additional Information

This appears to be caused by line 22 in multi-select.js - it uses a fixed id for the multiple select column

Whersi:
Closing this issue as a PR has been sent and merged into main https://github.com/ministryofjustice/moj-frontend/pull/930

PR #930, opened 2024-11-15
fix: (multi-select)  allow multiple multi-selects on same page

starswan:
### Identify the bug

https://github.com/ministryofjustice/moj-frontend/issues/914

### Description of the change

Add a optional 'idprefix' option to the multi-select component, so that the 'all' checkbox can have a unique id. Without this, multiple multi-selects on the same page don't work properly (selecting the second one 'all-selects' the first in Chrome, but YMMV)

### Alternative designs

None 

### Possible drawbacks

None - I have tried to ensure that the default id of the 'all' checkbox remains the same if the option is not provided.

### Verification process

This change is to support a feature in https://github.com/DFE-Digital/teaching-vacancies, where multiple tabs each have their own set of check-boxes. This change makes those check-boxes behave correctly and independently.

### Release notes

Fixed issue with multiple multi-selects appearing on the same page using new optional data-multi-select-idprefix.

### DIFF
--- a/docs/examples/multi-select/index.njk
+++ b/docs/examples/multi-select/index.njk
@@ -4,7 +4,7 @@ title: Multi select (example)
 figma_link: https://www.figma.com/file/N2xqOFkyehXwcD9DxU1gEq/MoJ-Figma-Kit?type=design&node-id=841-220&mode=design
 ---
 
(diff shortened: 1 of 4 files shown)


chrispymm (approved):
Thanks. This all looks good.

PR #976, opened 2024-12-02
docs(contributor template): add arguments to contributors partial

murrlipp:
- Add ability to distinguish contributed building blocks
- Remove redundant arguments for isIndex and isArchive
- Remove redundant component-no-feedback.njk

✅ Closes: 970


chrispymm (commented):
Are we certain that "contributing" and "giving feedback on" are the only valid responses? If we are then this is all good, but it feels like it may be limiting.

If not, then having contribution be a string and enabling a free text input for what the contribution is might work? Although this may result in repetitive entry of the 2 main options.

The third option would be to have what we have now with a third option if the `contribition` variable is a string

```
{% if contribution && contribution | is_string %}
  {{ contribiution }}
{% else if contribution %}
      contributing
    {% else %}
      giving feedback on
    {% endif %}
```
NB the above would require the creation of an `is_string` filter

murrlipp:
> Are we certain that "contributing" and "giving feedback on" are the only valid responses? If we are then this is all good, but it feels like it may be limiting.
> 
> If not, then having contribution be a string and enabling a free text input for what the contribution is might work? Although this may result in repetitive entry of the 2 main options.
> 
> The third option would be to have what we have now with a third option if the `contribition` variable is a string
> 
> ```
> {% if contribution && contribution | is_string %}
>   {{ contribiution }}
> {% else if contribution %}
>       contributing
>     {% else %}
>       giving feedback on
>     {% endif %}
> ```
> 
> NB the above would require the creation of an `is_string` filter

Spoke about this with Helen and we decided to cater for what we know now - components will either be contributed or given feedback on.  However, if this changes in future we can revisit.

I would imagine any future variations might require further changes than just this short string so we might need to rethink generally, if there do turn out to be scenarios outside of 'contributing' or 'giving feedback on' a given building block.

chrispymm (approved):

Issue #907, opened 2024-11-04
Figma library instructions


robertjmccarthy:
## What
Add instructions to the design system to help users add the MoJ Figma Kit. A good example can be seen on the SIS design system website: https://design.sis.gov.uk/get-started/figma 

## Why
It isn't necessarily easy or obvious as to how to add a Figma library for use in a Figma file, especially where some users may not be confident or highly skilled with Figma. The DS team want DS users to use the MoJ Figma Kit as a library as this means they'll benefit from updates. It's in the best interest of DS users and the DS team that the MoJ Figma Kit is added and used as a library.

## Who needs to be involved
- Interaction Designer
- Content Designer

## Done
When prototyping information on the DS website is updated to include information about Figma prototyping.

robertjmccarthy:
@murrlipp maybe this is something that you could look at with your workshop for the IxD CoP. Not sure if you can weave this in, or if there's a need to, but we currently don't have anything on the DS doc site about using the MoJ Figma Kit and how to use it as a library.

helennickols:
@robertjmccarthy @murrlipp Is this dependent on content design community involvement? If so, I reckon we can get a session in before Jan (an extra one to the regular CoP session.

murrlipp:
@helennickols I have published a branch with draft guidance on: https://github.com/ministryofjustice/moj-frontend/tree/add-figma-guidance

If you wouldn't mind pulling down and checking when you have time it would be appreciated.

helennickols:
I've started some changes for this. I'll pick them up with Murray on Friday. It's looking good.

helennickols:
Putting this in blocked until I can work on things locally.

helennickols:
Helen to come on to this on Monday.

helennickols:
IC is all in review so now picking this up

helennickols:
I have 2 conflicts in Github and so cannot view the branch. Murray has shared the Markdown, which I am looking at

helennickols:
---
layout: layouts/get-started.njk
subsection: Prototyping
title:  Using the Figma Kit
eleventyNavigation:
  key: Using the Figma Kit
  parent: Prototyping
  order: 30
  excerpt: "Using the MoJ Design System Figma Kit."
---

[...]

PR #978, opened 2024-12-03
docs: Add figma guidance

murrlipp:
Adding Figma guidance.


robertjmccarthy (approved):
Approving PR so it can be viewed on staging. @murrlipp and I have had an offline review and spoken about some changes which are documented on #907

PR #989, opened 2024-12-10
feat(tabbed-view-documents): tabbed view documentation

jbarnardmoj:
Ability to tab content in documentation.


.eleventy.js
```diff
@@ -130,6 +130,53 @@ module.exports = function (eleventyConfig) {
...
+    </div>
+  `.trim();
+  });
+
+  // Find and store govuk tab for above tabs
+  eleventyConfig.addPairedShortcode("tab", function (content, label) {
+    tabsStorage.push({ label, content });
+    return "";
+  });
+
```
chrispymm, line 179: I think at some point we should tidy up the .eleventy.js file - its getting a bit out of hand. Should probably extract things to their own file and import them. But not for this PR.

docs/assets/stylesheets/components/_documentation_tabs.scss
```diff
@@ -0,0 +1,281 @@
+.app-navigation {
```
chrispymm, line 1: The only thing I'm not a great fan of is this being app-_navigation_ 

I can see what we're trying to achieve, and that we're basically overlaying the navigation style over it, but it still feels a bit off.  I realise that the class `.app-tabs` is already taken for the examples... but I do sort of feel that this should be its own thing (or "block" I suppose in BEM terms)

We have 2 different tabs components:
 - example tabs
 - documentation/content tabs

To my mind they both have the same behaviours so should have the same js `data-module` (they currently don't - the example tabs use our own tabs js and the content ones use govuk), and then all the styles for them should be wrapped up into a block level class for that style of tabs. e.g. `.app-example-tabs` and `.app-content-tabs` as I don't imagine there are many shared styles so they're not really a variant.

What are your thoughts?

jbarnardmoj: Yes I do understand and agree. Although they behave slightly different - the document/content tabs turn into anchor links on mobile screens whilst the example tabs remain tabs. I think the idea was to utilise the tab component for functionality but we could build that logic separately.

.eleventy.js
```diff
@@ -132,6 +132,53 @@ module.exports = function (eleventyConfig) {
...
+    }).join("\n").trim();
+
+    const tabPanels = tabsStorage.map((tab, index) => {
+      const isHidden = index === 0 ? '' : ' govuk-tabs__panel--hidden';
+      return `
+      <div class="govuk-tabs__panel${isHidden}" id="${tabId(tab)}" role="tabpanel">
+        ${tab.content}
+      </div>
+    `.trim();
+    }).join("\n").trim();
```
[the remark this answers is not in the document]
jbarnardmoj: yes, good spot. Updated.

docs/assets/stylesheets/components/_documentation_tabs.scss
```diff
@@ -0,0 +1,284 @@
...
```
chrispymm, line 1: still not wild about this classname, but I can live with it.

However I think some of these customisations may need to also include the `.js-enabled` class, or some extra work needs to happen for the no-js version as it currently looks like this:

image

The gov.uk tabs look like this (without JS):
image

I think I'd err towards having it look as per gov.uk tabs without js and then enhance to what we have now. Might be as simple as `.js-enabled .app-navigation`

chrispymm (approved):

PR #1004, opened 2024-12-13
Organised eleventy

jbarnardmoj:
# Pull-request templates

Please follow the steps below to tell us about your contribution.

1. Copy the correct template for your contribution

- Are you fixing a bug? [Copy the template](https://github.com/ministryofjustice/moj-frontend/tree/main/.github/PULL_REQUEST_TEMPLATE/BUG_FIX.md)
- Are you improving performance? [Copy the template](https://github.com/ministryofjustice/moj-frontend/tree/main/.github/PULL_REQUEST_TEMPLATE/PERFORMANCE_IMPROVEMENT.md)
- Are you updating documentation? [Copy the template](https://github.com/ministryofjustice/moj-frontend/tree/main/.github/PULL_REQUEST_TEMPLATE/DOCUMENTATION.md)
- Are you changing functionality? [Copy the template](https://github.com/ministryofjustice/moj-frontend/tree/main/.github/PULL_REQUEST_TEMPLATE/FEATURE_CHANGE.md)

2. Replace this text with the contents of the template
3. Fill in all sections of the template
4. Click "Create pull request".

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -147,52 +147,12 @@ module.exports = function (eleventyConfig) {
     `;
   });
 
-  eleventyConfig.addFilter(
-    "addActiveAttribute",
-    function (config, filePathStem) {
-      if (config.items) {
-        return {
-          ...config,
-          items: config.items.map((item) => ({
-            ...item,
-            active: filePathStem.indexOf(item.href) > -1,
-          })),
-        };
-      } else if (config.sections) {
-        return {
-          ...config,
-          sections: config.sections.map((section) => ({
-            ...section,
-            items: section.items.map((item) => ({
-              ...item,
-              active: filePathStem.indexOf(item.href) > -1,
-            })),
-          })),
-        };
-      }
-    },
-  );
-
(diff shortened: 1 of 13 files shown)


chrispymm:
I think this PR goes a bit further than I anticipated, and IMO ends up with a slightly unreadable .eleventy.js file - it feels like it ends up being hard to scan that file and see what filters / shortcodes are enabled and most importantly the string I need to use to invoke them, which isn't ideal.

I think I was expecting each extracted filter/shortcode to export the filter/shortcode function, and then the eleventy config file would still have the addShortcode definition bit

```
const tabs = require('shortcodes/tabs.js')

eleventyConfig.addPairedShortcode("tabs", tabs)
```

This extracts away the implementation, but leaves someone able to read the eleventy config file and see what shortcodes and filters are available to them, and how to invoke them.

This may cause scoping issues for the tabs / tab shortcode though...

I think if we can't do the above I'd still prefer manually requiring each file within .eleventy.js rather than looping through a directory

PR #1092, opened 2025-01-20
docs: add google tag manager

murrlipp:


robertjmccarthy (commented):
The existing Google Tag Manager code that you've deleted is to track analytics. I've just checked the ID and it matches what is in our Google Analytics.

So we don't want to delete that otherwise we lose the connection.

chrispymm (approved):
Based on slack conversations, this is good to go!

Issue #1107, opened 2025-01-22
Sortable table: Using parseFloat to evaluate cell values removes parts of values if there is a number prefix


natclamp-moj:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of the following:
  - Reproduced the problem
  - Followed all applicable steps in the debugging guide
  - Checked the FAQs on the message board for common solutions
  - Checked that your issue doesn't exist: https://github.com/ministryofjustice/moj-frontend/issues
  - Checked that there is not already a package that provides the described functionality

### Description

The getCellValue function includes a parseFloat which removes any non-number after a number within a value (e.g. '123ABC' becomes '123') which means the sorting of these values is impossible.

### Steps to Reproduce

1. Use the sortable table component and have data of different types in one column, such as number, string with number prefix, string, empty string, etc.
2. Sort the table using this column

**Expected behaviour:** If you log out the cell values, any cells with strings beginning with numbers (e.g. '123A') will have been converted to 123 because of the parseFloat used in the getCellValue function. I expect sorting to be on the entire cell value, e.g. '123A' before '123B'.

**Actual behaviour:** If you have multiple similar values, this means they won't sort correctly (e.g. '123A' and '123B' would just be 123 vs 123)

**Reproduces how often:** 100%

### Versions

[3.3.1](2025-01-09)
OS: Mac Sequoia 15.2
Browser: chrome

PR #1108, opened 2025-01-22
SAA-2291: Introduce convention for strings with number prefix

natclamp-moj:
### Identify the bug

https://github.com/ministryofjustice/moj-frontend/issues/1107

### Description of the change

Instead of running all cell values through `parseFloat`, which converts strings with number prefixes to just the number, we use `Number` to evaluate the entire value to see whether it's a number or not. If the value is NaN, the function returns the original value (this includes strings with number prefixes) for use in the sort. If the value is a number, we can then safely use `parseFloat` on the sort value, as we know it won't be removing any non-numerical components of the value.

### Alternative designs

`Number` evaluates the entire value, whereas `parseFloat` stops when you get to a non-numerical part. 
You could use `parseFloat` and then check the result against the original input value, and use the original value if it is different from the result of the `parseFloat`, but this is more long-winded and potentially error-prone.

### Possible drawbacks

`Number` converts empty strings and `null` to 0. I think that this works fine when sorting - if I had a cell which looked empty, I'd rather it was at the bottom of the table rather than at the top - but this depends on your point of view.

### Verification process

Verified against existing tests and introduced a new test with a variety of input values.

### Release notes

- Provides a better convention for sorting if strings contain a number prefix


chrispymm (approved):

src/moj/components/sortable-table/sortable-table.js
```diff
@@ -118,7 +118,7 @@ MOJFrontend.SortableTable.prototype.sort = function(rows, columnNumber, sortDire
 
 MOJFrontend.SortableTable.prototype.getCellValue = function(cell) {
 	const val = cell.attr('data-sort-value') || cell.html();
-
-	const floatVal = parseFloat(val)
-	return isNaN(floatVal) ? val : floatVal
+	
+	const valAsNumber = Number(val)
+	return isNaN(valAsNumber) ? val : parseFloat(val)
```
jbarnardmoj, line 123: parseFloat isn't needed here as we already have a number (which might be a float) - parseFloat parses a string.

jbarnardmoj (changes requested):

chrispymm (approved):

jbarnardmoj (approved):

PR #1119, opened 2025-01-27
style: add prettier code formatting

chrispymm:
This PR installs and configures prettier, and applies it to all js, css, scss, json, and yml files.


.editorconfig
```diff
@@ -1,4 +1,4 @@
-# http://EditorConfig.org
+# http://EditorConfig.or# http://EditorConfig.org
```
colinrotherham, line 1: Ah is this an accidental typo?

chrispymm: yep, cop/paste error 🤦

.prettierignore
```diff
@@ -0,0 +1,18 @@
+# Temporary only
+.cache/
+.tmp/
+
+**/*.md
```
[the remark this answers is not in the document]
chrispymm: yeah, I wasn't 100% confident what it was gonna do to our markdown files, so I thought I'd address that in a separate PR that was a bit more focused.

colinrotherham (approved):

Issue #1123, opened 2025-01-28
Document component JavaScript using JSDoc


colinrotherham:
## What
Add JSDoc comments to all component JavaScript including ESLint and compiler checks

## Why
As part of the work to modernise our codebase, we'd like to follow GOV.UK Frontend by adding [JSDoc comments](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) to better document component JavaScript methods, options and properties.

This will include work to mark component methods as "public" or "private".

**Related:** See [GOV.UK Frontend generated JSDoc output](https://govuk-frontend-review.herokuapp.com/docs/javascript/) for reference

## Who needs to be involved
- Developer

## Done
1. Align ESLint with GOV.UK Frontend
2. JSDoc added to component JavaScript methods, options and properties
3. Linting task `lint:js` updated to check JSDoc comments
4. Linting task `lint:types` added to validate JSDoc comments

PR #1129, opened 2025-01-30
style: align ESLint with GOV.UK Frontend

colinrotherham:
Aligns ESLint configs and fixes various flagged issues as part of https://github.com/ministryofjustice/moj-frontend/issues/1123

### DIFF
--- a/src/moj/components/date-picker/date-picker.spec.js
+++ b/src/moj/components/date-picker/date-picker.spec.js
@@ -780,7 +780,7 @@ describe('button menu JS API', () => {
       expect(input).toHaveValue(dateToSelect.format('DD/MM/YYYY'))
     })
 
-    test.skip.failing('minDate', async () => {
+    test('minDate', async () => {
       const minDay = 3
       const lastDayinMonth = dayjs().endOf('month').date()
       const minDate = dayjs().date(minDay)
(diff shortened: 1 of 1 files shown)


src/moj/components/multi-file-upload/multi-file-upload.js
```diff
@@ -199,7 +199,7 @@ if (
       url: this.params.deleteUrl,
       type: 'post',
       dataType: 'json',
-      data: data,
+      data,
```
chrispymm, line 202: Very much a personal preference but I don't like this style at all! It's not saving any typing, and to me it feels less obvious what is happening, especially when it is mixed in with other `key: value` pairs. But I accept that part of this whole process is removing personal preference infavour of consistent/enforced code style.

.eslintrc.js
```diff
@@ -130,11 +139,15 @@ module.exports = {
       processor: 'markdown/markdown'
     },
     {
-      // Check markdown `*.md` contains valid code blocks
-      // https://www.npmjs.com/package/eslint-plugin-markdown#user-content-advanced-configuration
-      files: ['**/*.md/*.{cjs,js,mjs}'],
+      files: [
+        '**/assets/javascript/**/*.{cjs,js,mjs}',
+        '**/docs/examples/**/*.{cjs,js,mjs}',
+        '**/*.md/*.{cjs,js,mjs}'
```
chrispymm, line 145: Why do we have the `md` check files line up top and down here? I think maybe I don't fully understand how each of these override blocks are applied / relate to each other...

chrispymm (approved):

Issue #1141, opened 2025-02-05
Update JavaScript build to bundle modules (versus concatenation)


colinrotherham:
## What

Update published JavaScript build to use [`rollup`](https://rollupjs.org) to match GOV.UK Frontend

## Why

We're currently using various Gulp plugins for our published JavaScript

Due to how scripts are concatenated we can't currently bundle `module.exports` or in future `export`

## Who needs to be involved

- Developer

## Done

- Gulp packages updated
- Build packages updated
- JavaScript builds use `rollup`

PR #1149, opened 2025-02-06
feat: migrate to bundled modules using Rollup

colinrotherham:
This PR updates all component JavaScript with:

1. CommonJS `module.exports` in source code
2. Universal Module Definition (UMD) when packaged

~We may need to switch esbuild to Rollup as UMD bundle support isn't included out of the box~

Part of https://github.com/ministryofjustice/moj-frontend/issues/1141

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -57,7 +57,7 @@ gulp.task('watch:package-js', () => {
 // Watch the docs js files and the bundled package js and rebuild
 gulp.task('watch:docs-js', () => {
   gulp.watch(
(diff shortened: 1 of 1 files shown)


src/moj/components/date-picker/date-picker.js
```diff
@@ -1,7 +1,7 @@
 /**
- * Datepicker config
+ * Daste picker config
```
chrispymm, line 2: Typo!

chrispymm (approved):

PR #1180, opened 2025-02-24
fix: various issues spotted by `tsc` TypeScript compiler

colinrotherham:
This PR fixes various issues spotted by the TypeScript compiler

These were flagged by `npm run lint:types` as part of #1123 in:

* https://github.com/ministryofjustice/moj-frontend/pull/1128

**Note:** Be aware that some of these fixes are a bit pedantic

### DIFF
--- a/src/moj/components/multi-file-upload/multi-file-upload.spec.mjs
+++ b/src/moj/components/multi-file-upload/multi-file-upload.spec.mjs
@@ -156,20 +156,24 @@ describe('Multi-file upload', () => {
     })
 
     test('displays upload progress', async () => {
+      /** @type {SinonFakeXMLHttpRequest | undefined} */
+      let request
+
       // Create a spy on XMLHttpRequest to simulate upload progress
       const xhr = useFakeXMLHttpRequest()
-      let request
       xhr.onCreate = (req) => {
         request = req
       }
 
       await user.upload(input, file)
 
-      request.uploadProgress({
-        lengthComputable: true,
-        loaded: 50,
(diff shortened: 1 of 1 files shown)


src/moj/filters/all.js
```diff
@@ -60,8 +73,17 @@ module.exports = function () {
...
-  ------------------------------------------------------------------ */
-  return filters
+  /**
+   * Returns an object used to store the methods registered as a
+   * 'filter' (of the same name) within nunjucks. You can override
+   * gov.uk core filters by creating filter methods of the same name.
+   */
+  return {
+    date,
+    mojDate
```
chrispymm, line 83: Could the fact we are not exporting the other functions cause anyone a problem if they are using them directly in prototypes? Or are these functions suplicates of ones already available within the prototype kit?

src/moj/components/sortable-table/sortable-table.mjs
```diff
@@ -136,27 +148,33 @@ SortableTable.prototype.getTableRowsArray = function () {
...
-    }.bind(this)
-  )
-  return newRows
+  return rows.sort((rowA, rowB) => {
+    const tdA = rowA.querySelectorAll('td, th')[columnNumber]
+    const tdB = rowB.querySelectorAll('td, th')[columnNumber]
+
+    if (
+      !tdA ||
+      !tdA ||
```
chrispymm, line 157: Think this should be `!tdB`

chrispymm (approved):

PR #1193, opened 2025-02-26
docs(side nav): update side nav to remove css :has for better browser support

chrispymm:
Removes the use of CSS `:has` from within the side nav JS.

### DIFF
--- a/docs/assets/javascript/collapsible-nav.mjs
+++ b/docs/assets/javascript/collapsible-nav.mjs
@@ -5,13 +5,14 @@ export default class CollapsibleNav extends HTMLElement {
       this.id = `nav-${Date.now()}`
     }
 
-    this.collapsibleItems = this.querySelectorAll('li:has(ul)')
-
     this.collapsibleItems.forEach(($item, index) => {
       const $link = $item.querySelector('a')
       const $list = $item.querySelector('ul')
-      const $button = document.createElement('button')
 
+      if (!$list) return
+      if (!$link) return
+
+      const $button = document.createElement('button')
       if (!$list.id) {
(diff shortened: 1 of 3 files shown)


docs/assets/javascript/collapsible-nav.mjs
```diff
@@ -83,6 +84,14 @@ export default class CollapsibleNav extends HTMLElement {
   }
 
   get openItems() {
-    return this.querySelectorAll(`li:has(ul).${this.openClass}`)
+    return this.collapsibleItems.filter((el) => {
```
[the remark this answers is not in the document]
chrispymm: deploying preview now

docs/assets/javascript/collapsible-nav.mjs
ehtabrizi, line 1: Only minor, should a lifecycle hook 'connectedCallback()' be used to ensure the element is in the DOM?

chrispymm: Yeah, this is an interesting one. I've seen webcomponents configured both ways, and conflicting arguments for best practice.

Ultimately here, I don't think it makes a difference, our code is in a `<script type="module">` so is `defer` by default.  This means the element should always be present in the document at the time the JS is run.

ehtabrizi: Sure, wasn't sure if the menu is added or removed at any point. Since it's static, my comment shouldn't be needed. Thought I'd best check, thanks Chris.

chrispymm: yeah, it nerver gets removed / added into the DOM

PR #1249, opened 2025-03-07
docs: fix example require path to Nunjucks filters

colinrotherham:
As mentioned on Slack, it's not clear what the correct filters `require()` path is

This PR fixes the path typo ~`frontend/filters/all`~ `frontend/moj/filters/all`

```patch
- require('./node_modules/@ministryofjustice/frontend/filters/all')
+ require('@ministryofjustice/frontend/moj/filters/all')
```

I've suggested improved (but generic) instructions here:

---

### Installation

You will need to install the MoJ Frontend Nunjucks filters to use the `mojTimeline()` Nunjucks macro:

1. Import the filters

   ```mjs
   import mojFilters from '@ministryofjustice/frontend/moj/filters/all.js'
   ```

2. Add the filters to your Nunjucks environment

   ```mjs
   for (const [name, filter] of Object.entries(mojFilters())) {
     nunjucksAppEnv.addFilter(name, filter)
   }
   ```

### DIFF
--- a/src/moj/components/messages/README.md
+++ b/src/moj/components/messages/README.md
@@ -4,16 +4,21 @@
 
 ### Installation
 
-You will need to install the following code at the bottom of `server.js`, just above `module.exports = app`
+You will need to install the MoJ Frontend Nunjucks filters to use the `mojMessages()` Nunjucks macro:
 
-```js
-const mojFilters = require('@ministryofjustice/frontend/moj/filters/all')
+1. Import the filters
 
-// Add filters from MoJ Frontend
-for (const [name, filter] of Object.entries(mojFilters())) {
-  nunjucksAppEnv.addFilter(name, filter)
-}
-```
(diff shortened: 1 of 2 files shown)


src/moj/components/messages/README.md
```diff
@@ -4,15 +4,15 @@
 
 ### Installation
 
-You will need to install the following code at the bottom of `server.js`, just above `module.exports = app;`
+You will need to install the following code at the bottom of `server.js`, just above `module.exports = app`
```
chrispymm, line 7: What is server.js? It's not part of our code... 

I assume it is [this file](https://github.com/alphagov/govuk-prototype-kit/blob/main/server.js) from the prototype kit, but that's not obvious, and not everyone uses the prototype kit, so we probably need some clarification here.

chrispymm (approved):

Issue #901, opened 2024-10-30
Remove dependency on jQuery from all components


chrispymm:
## What
some of our components still depend on the 3rd party jQuery library.  In modern browsers with modern JS jQuery is no longer necessary. 

# Why
To bring us into line with the GOV.UK design system (they dropped jQuery in 2023). To modernise our code, and to allow services using our components to not have to include jQuery - improving their performance. 

## Who needs to be involved
- Developer

## Done

- [ ] Add another
- [ ] Filter Toggle
- [ ] Multi select
- [ ] Password Reveal
- [ ] Search Toggle
- [ ] Multi File upload
- [ ] Sortable table

robertjmccarthy:
Closing ticket #134 from backlog as duplication with this ticket. There are comments on #134 that may be useful as a reference for this work.

PR #1283, opened 2025-03-18
feat: remove jQuery

colinrotherham:
Closes https://github.com/ministryofjustice/moj-frontend/issues/901 😱

### DIFF
--- a/src/moj/components/button-menu/README.md
+++ b/src/moj/components/button-menu/README.md
@@ -28,12 +28,16 @@ The macro
 
 The JavaScript
 
-```js
-new MOJFrontend.ButtonGroup({
-  container: document.querySelector('.moj-button-group'),
-  mq: '(min-width: 45em)',
-  buttonText: 'Actions'
-});
+```mjs
+import { ButtonMenu } from '@ministryofjustice/frontend'
+
+const $buttonMenu = document.querySelector('.app-button-menu')
+
+new ButtonMenu($buttonMenu, {
+  alignMenu: 'right',
+  buttonText: 'Menu title',
+  buttonClasses: 'govuk-button--secondary'
+})
 ```
 
 ## Arguments
--- a/docs/components/multi-file-upload.md
+++ b/docs/components/multi-file-upload.md
@@ -46,16 +46,18 @@ When a file has been uploaded it will show as:
 
 ### Initialising the JavaScript
 
-The multi file upload component uses JavaScript. To run it you must include the following script in your page:
+The multi file upload component uses JavaScript. To run it you must include the following:
 
-```js
-if (typeof MOJFrontend.MultiFileUpload !== 'undefined') {
-  new MOJFrontend.MultiFileUpload({
-    container: document.querySelector('.moj-multi-file-upload'),
-    uploadUrl: '/ajax-upload-url',
-    deleteUrl: '/ajax-delete-url'
-  });
-}
+```mjs
+import { MultiFileUpload } from '@ministryofjustice/frontend'
+
+const $multiFileUpload = document.querySelector('.app-multi-file-upload')
+
+new MultiFileUpload({
+  container: $multiFileUpload,
+  uploadUrl: '/ajax-upload-url',
+  deleteUrl: '/ajax-delete-url'
+})
 ```
 
 ### When JavaScript is not available
--- a/src/moj/components/rich-text-editor/README.md
+++ b/src/moj/components/rich-text-editor/README.md
@@ -4,13 +4,19 @@
 
 ## Example
 
-```html
-<script type="module">
-  new MOJFrontend.RichTextEditor({
-    textarea: document.querySelector('.js-editor'),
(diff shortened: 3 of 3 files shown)


src/moj/components/button-menu/README.md
```diff
@@ -30,7 +30,7 @@ The JavaScript
 
 ```js
 new MOJFrontend.ButtonGroup({
-  container: $('.moj-button-group'),
+  container: document.querySelector('.moj-button-group'),
   mq: '(min-width: 45em)',
   buttonText: 'Actions'
 });
```
chrispymm, line 36: Pretty sure this can be removed as it's incorrect! `new MOJFrontend.ButtonGroup` doesn't actually exist! This should likely have been removed when we updated the button menu component - the nunjucks and the js example are wrong. At a glance the args look correct

src/moj/components/sortable-table/sortable-table.mjs
```diff
@@ -25,71 +28,93 @@ SortableTable.prototype.setupOptions = function (params) {
...
-      const sortedRows = this.sort(rows, columnNumber, sortDirection)
+  for (const heading of this.headings) {
+    const button = heading.querySelector('button')
+    const sortDirection = heading.getAttribute('aria-sort')
+
+    if (sortDirection === 'ascending' || sortDirection === 'descending') {
+      const columnNumber = button.getAttribute('data-index')
+      const sortedRows = this.sort(columnNumber, sortDirection)
+
       this.addRows(sortedRows)
```
chrispymm, line 70: The jquery version has a `.first()` so it only applies an initial sort to the first heading it finds with a direction.  The new version goes through all headings applying a sort to each, so in a table with multiple columns having an initial sort direction applied, the new behaviour would be different - as it would be the last column that was sorted, not the first.

We probably want to break out of the `for` loop once we find a heading with a `sortDirection`

src/moj/components/multi-file-upload/multi-file-upload.mjs
```diff
@@ -12,81 +10,89 @@ export function MultiFileUpload(params) {
...
-  this.dropzone.on('drop', this.onDrop.bind(this))
+  this.dropzone = document.createElement('div')
+  this.dropzone.classList.add('moj-multi-file-upload__dropzone')
+
+  this.dropzone.addEventListener('dragover', this.onDragOver.bind(this))
+  this.dropzone.addEventListener('dragleave', this.onDragLeave.bind(this))
+  this.dropzone.addEventListener('drop', this.onDrop.bind(this))
+
+  this.fileInput.replaceWith(this.dropzone)
+  this.dropzone.appendChild(this.fileInput)
```
chrispymm, line 48: Does this work in this order? I assume it you've checked and it does, but on first reading it doesn't seem like it should! If fileInput has been replacedWith can it be appended?

src/moj/components/multi-file-upload/multi-file-upload.mjs
```diff
@@ -12,81 +10,89 @@ export function MultiFileUpload(params) {
...
 MultiFileUpload.prototype.onDrop = function (event) {
   event.preventDefault()
-  this.dropzone.removeClass('moj-multi-file-upload--dragover')
-  this.feedbackContainer.removeClass('moj-hidden')
-  this.status.html(this.params.uploadStatusText)
-  this.uploadFiles(event.originalEvent.dataTransfer.files)
+  this.dropzone.classList.remove('moj-multi-file-upload--dragover')
+  this.feedbackContainer.classList.remove('moj-hidden')
+  this.status.textContent = this.params.uploadStatusText
+  this.uploadFiles(event.dataTransfer.files)
```
chrispymm, line 95: Are we sure `originalEvent` isn't required here?

chrispymm: Ah yes, I forget that you get a jQuery modified event!

src/moj/all.mjs
```diff
@@ -133,6 +132,7 @@ export {
   ButtonMenu,
   DatePicker,
   FilterToggleButton,
+  FormValidator,
```
chrispymm, line 135: I'm not sure this should be added back in... it is archived...

But then so is richTextEditor and passwordReveal... I suspect that they should have been removed when they were archived, but can't say for certain as we don't really have a process/definition for archived components and what happens to them ☹️

src/moj/components/button-menu/README.md
chrispymm: Yeah might be worth a separate PR - its not critical, I'm not sure how many people read the component READMEs anyway, some of them don't even have one!

src/moj/all.mjs
chrispymm: Ah ok, that's fine if its a fix for something we've broken along the way. That makes sense. In which case lets leave it in here.  We'll deal with the archived-but-still-actually-there components another day - we should probably warn people we're going to deprecate them first, and then remove them later

chrispymm (approved):
🚀

PR #1328, opened 2025-03-25
feat: default JavaScript bundles to ES modules

colinrotherham:
This PR changes the default JavaScript bundles from UMD (Universal Module Definition) to ES Modules

* [GOV.UK Frontend **JavaScript browser compatibility**](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/006-javascript-compatibility.md)
* [GOV.UK Frontend **JavaScript file formats**](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/007-javascript-file-formats.md)

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -367,7 +367,7 @@ module.exports = function (eleventyConfig) {
     domDiff: false,
     port: 8080,
     watch: [
-      'public/javascripts/application.js',
+      'public/javascripts/application.bundle.js',
       'public/javascripts/govuk-frontend.min.js',
       'public/javascripts/moj-frontend.min.js',
       'public/stylesheets/application.css',
--- a/docs/_includes/layouts/base.njk
+++ b/docs/_includes/layouts/base.njk
@@ -54,6 +54,6 @@
 {% endblock %}
 
 {% block bodyEnd %}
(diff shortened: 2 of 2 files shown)


gulp/dist.js
```diff
@@ -1,3 +1,4 @@
+const { version } = require('@ministryofjustice/frontend/package.json')
```
chrispymm, line 1: Will this be correct? At the time of running this script, the version will be the _current_ version, and then once the script has built, then semantic release will run, which will then update the version in the published package?

I haven't got to a solution for this yet, as its a bit of a pain.

gulp/tasks/scripts.js
```diff
@@ -7,6 +7,8 @@ const terser = require('@rollup/plugin-terser')
 const PluginError = require('plugin-error')
 const { rollup } = require('rollup')
 
+const MOJFrontend = require('../../src/moj/all.mjs')
+
 /**
  * Compile Sass to CSS task
```
chrispymm, line 13: Is this a copy/paste error? As this function has nothing to do with sass/css - not a big deal, just confused me for a moment or two!

chrispymm (approved):

PR #1329, opened 2025-03-26
docs: style guide correction

helennickols:


helennickols:
> > Yes. Is that behaviour in place for other entries on the updates page?
> 
> It's automatic but only in Google Chrome and Edge
> 
> Other browsers (e.g. Safari, Firefox) just go to the top of the page

Yes please. Could we do this for all entries? Is it a lot of work? If so, let's ship this first.

PR #1430, opened 2025-05-12
Remove accessibility report uploads

chrispymm:
This PR removes the accessibility report uploads and makes the issues discovered fields for the accessiblity sections required.

This is due to accessibility reports not being publishable to the public domain, but we still need to know the findings from the audits done.

### DIFF
--- a/schema/add-assistive-tech.schema.js
+++ b/schema/add-assistive-tech.schema.js
@@ -76,8 +76,8 @@ const schema = Joi.object({
     .required()
     .custom((value, helpers) => maxWords(value, helpers, 250))
     .messages({
-      'any.required': 'Enter any issues discovered',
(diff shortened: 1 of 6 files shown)


schema/add-assistive-tech.schema.js
```diff
@@ -73,20 +73,13 @@ const schema = Joi.object({
     }),
 
   issuesDiscovered: Joi.string()
-    .optional()
-    .allow(null, '')
+    .required()
     .custom((value, helpers) => maxWords(value, helpers, 250))
     .messages({
+      'any.required': 'Enter any issues discovered',
+      'string.empty': 'Enter any issues discovered',
```
chrispymm, line 80: @helennickols You will likely want to alter these validation messages that are needed now the field is required.

PR #1543, opened 2025-06-23
docs: adding PDU entry

helennickols:


views/common/homepage-nav.html
murrlipp, line 1: I think this change has happened as a result of some local/working directory weirdness.

It doesn't affect the other work you've done but probably best to change it back.

views/common/homepage-nav.html
```diff
@@ -403,7 +403,7 @@ <h2 class="app-vertical-nav__header" id="guidance-nav-title">Guidance</h2>
   <a class="app-vertical-nav__link" href="/prototyping/setting-up-figma-prototypes/">Setting up Figma prototypes</a></li>
 
 
-
+
```
murrlipp, line 406:
```suggestion
```

PR #1547, opened 2025-06-25
404 and 500 pages

chrispymm:
This PR adds in 404 and 500 error pages.  
Nunjucks templates are copies from 11ty to the express app 
Nginx is configured to serve the static pages generated by 11ty, and then express handles dynamic errors when in the contributions app


helennickols:
@chrispymm Can I do this or does it need a dev?

PR #1741, opened 2025-09-03
WIP: Refactor content style guide

chrispymm:
Builds on the work in simplify content style guide to refactor the style guide page itself.

Each style guide entry becomes its own .md file 

```
---
title: entry title
---
Entry content markdown
```

We then use an 11tydata file to assign a 'tag' to all the entries to group them into an eleventy collection, and also use `eleventyComputed` to generate a a data entry for the first letter of the title

The style guide layout file can then wrangle the data from the collection into a format that can be passed into the govuk-accordion nunjucks macro.

This hugely simplifies the addition and editing of entries in the style guide, as we will now only be dealing directly with the content in .md files, not a big mess of HTML.


chrispymm:
Closed and replaced with #1847

PR #1779, opened 2025-09-08
New component: PDS Header

ms-ek:
### Description of the change

This PR introduces a new **PDS Header** (`pds-header`) component, created by the Probation team within MOJ.

- The component is intended **only for Probation services** and is primarily to be used by **designers for prototyping purposes**.
- Developers should continue to follow the implementation guidance in the Probation components NPM package: <https://www.npmjs.com/package/@ministryofjustice/hmpps-probation-frontend-components>.
- SCSS is using scoped selectors to avoid impacting other GOV.UK/MoJ components.
- Includes **unit tests** for initialisation and user menu toggling.
- Manually integration-tested via `npm link` in **hmpps-template-typescript**: <https://github.com/ministryofjustice/hmpps-template-typescript>.

> Note: This addition was not explicitly requested by the moj-frontend team; it is proposed to support ongoing Probation work across multiple services.

---

### Alternative designs

- **Keep the header only in `@ministryofjustice/hmpps-probation-frontend-components` (no change here)**
  - **Pros:** Keeps `moj-frontend` narrowly scoped.
  - **Cons (main one):** Designers would need to follow the **developer setup** to consume the component, which may **not be possible in some project setups**. This increases adoption friction and risks inconsistent implementations.

- **Provide guidance/snippets to extend the GOV.UK header (no packaged component)**
  - **Pros:** Minimal footprint in this repo.
  - **Cons:** Inconsistent implementations; more copy-paste; harder to maintain parity with design.

- **Proposed approach (this PR):** Host a scoped, Probation-only component in `moj-frontend`.
  - **Pros:** Single upstream source for designers; lowers adoption barriers; reduces duplication.
  - **Cons:** Expands repo surface area with a domain-specific component.

---

### Possible drawbacks

- **Scope creep:** Introduces a domain-specific component into a shared library.
- **Governance:** Needs clear documentation that it’s **Probation-only**, not a general MoJ component.

---

### Verification process

**Automated tests**
- Added unit tests for:
  - Component initialisation (including guard against double-initialisation).
  - User menu toggle behaviour (`aria-expanded`, `hidden`).

**Manual integration test (`npm link`)**
1. In this repo/branch: `npm install` → `npm run build` → `npm link`
2. In **hmpps-template-typescript**: `npm link @ministryofjustice/frontend`
3. Import and initialise the component on a sample page.
4. Verified:
   - Header and footer render as expected.
   - Toggling account menu updates `aria-expanded` and `hidden`.
   - Scoped styles do not affect unrelated components.

---

### Release notes

Add **PDS Header** (`pds-header`) as a scoped, Probation-only component for designers, with basic JS behaviour and unit tests. Developers should continue to use the guidance in `@ministryofjustice/hmpps-probation-frontend-components`.


chrispymm (commented):
This all looks good @ms-ek. 

One thought though, what was your logic for the directory being '_probation'? I guess to distinguish it from all the other component directories...

I'm just wondering whether to be a bit more future-proof, and to tie in with the terminology we use elsewhere, whether something like

'src/moj/components/domain-specific/probation'

Thoughts?

ms-ek:
> This all looks good @ms-ek.
> 
> One thought though, what was your logic for the directory being '_probation'? I guess to distinguish it from all the other component directories...
> 
> I'm just wondering whether to be a bit more future-proof, and to tie in with the terminology we use elsewhere, whether something like
> 
> 'src/moj/components/domain-specific/probation'
> 
> Thoughts?

Very good idea @chrispymm. I was trying to make it standout from the other components and make it clear that this is under the probation space. Added `_` in front so that if other domains are added in the future they all show at the top instead of being mixed with other components. Happy with `domain-specific/...` it makes it even clearer. Just made the change

Issue #1922, opened 2025-10-20
[Bug] MultiFileUpload: InvalidStateError when accessing xhr.responseText with responseType = 'json'


espedzhev:
### **Description**

The `MultiFileUpload` component throws an `InvalidStateError` when uploading or deleting files.  
The error occurs because the component sets `xhr.responseType = 'json'` but still tries to read `xhr.responseText`, which is invalid when the response type is not `'text'` or `''`.

This appears to be a regression introduced when [replacing jQuery’s `$.ajax` with native `XMLHttpRequest`](https://github.com/ministryofjustice/moj-frontend/commit/c4eb082e8afcf7720b008cf671ffe020ed22eb4a).  
In jQuery, `responseText` was always available (and contained the raw JSON string), but with native XHR, accessing it after setting `responseType = 'json'` causes an exception.

---

### **Steps to Reproduce**

1. Visit the official component example: [Multi-file upload – MOJ Design System](https://design-patterns.service.justice.gov.uk/components/multi-file-upload/)
2. Open your browser’s developer consol
3. Upload any file

**Expected behaviour:**  
File uploads complete successfully without errors.

**Actual behaviour:**  
An exception is thrown in the browser console:
```js
[Uncaught InvalidStateError: Failed to read the 'responseText' property from 'XMLHttpRequest':
The value is only accessible if the object's 'responseType' is '' or 'text' (was 'json').](multi-file-upload.mjs:241)
```

**Reproduces how often:** 100%

---

### **Versions**

- **MOJ Frontend:** 6.0.0  

---

### **Additional Information**

**Root cause:**  
The following lines incorrectly reference `xhr.responseText`:
```js
// src/moj/components/multi-file-upload/multi-file-upload.mjs
this.params.uploadFileExitHook(this, file, xhr, xhr.responseText)
this.params.uploadFileErrorHook(this, file, xhr, xhr.responseText, error)
this.params.fileDeleteHook(this, undefined, xhr, xhr.responseText)
```

Suggested fix: Replace `xhr.responseText` with `xhr.response`.

For example:
```patch
- this.params.uploadFileExitHook(this, file, xhr, xhr.responseText)
+ this.params.uploadFileExitHook(this, file, xhr, xhr.response)
```
I tested this change locally, and it resolves the issue.
The error is also reproducible on the official MOJ Design System component page linked above.

chrispymm:
[...]

PR #1932, opened 2025-10-23
fix(multifileupload): remove invalid calls to xhr.responseText (#1922)

chrispymm:
Currently the calls to the callback functions use `xhr.responseText` which throws an error due to the `responseType` being 'json'.  This error was introduced when migrating from jQuery which always has `responseText` available.  Looking at the type definitions, and the parameter in v4.02 before the migration from jQuery (https://github.com/ministryofjustice/moj-frontend/blob/1fb9f79d123096cc5d49f65bb5c49de71bb5fe9b/src/moj/components/multi-file-upload/multi-file-upload.mjs#L172) it appears that this argument was intended to be `statusText`.

### DIFF
--- a/src/moj/components/multi-file-upload/multi-file-upload.mjs
+++ b/src/moj/components/multi-file-upload/multi-file-upload.mjs
@@ -225,7 +225,7 @@ export class MultiFileUpload extends ConfigurableComponent {
       this.$status.textContent = xhr.response.success.messageText
 
       $actions.append(this.getDeleteButton(xhr.response.file))
-      this.config.hooks.exitHook(this, file, xhr, xhr.responseText)
+      this.config.hooks.exitHook(this, file, xhr, xhr.statusText)
     }
 
     const onError = () => {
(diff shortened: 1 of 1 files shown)


src/moj/components/multi-file-upload/multi-file-upload.mjs
```diff
@@ -238,7 +238,7 @@ export class MultiFileUpload extends ConfigurableComponent {
       $message.innerHTML = this.getErrorHtml(error)
       this.$status.textContent = error.message
 
-      this.config.hooks.errorHook(this, file, xhr, xhr.responseText, error)
+      this.config.hooks.errorHook(this, file, xhr, xhr.statusText, error)
```
espedzhev: This works, but I can’t help wondering why this argument is even needed. Could we just remove the unused parameter altogether instead?

espedzhev: Got it, that makes sense, thanks for clarifying. If you do decide to deprecate or rework the component later, it won’t be too hard for us to replace it on our side.

PR #1952, opened 2025-11-03
Fast follows: PDS Header and Footer

ms-ek:


docs/_includes/example.njk
```diff
@@ -25,8 +25,14 @@
...
-    {%- if htmlWarning -%}
-      <div class="app-example__warning">{{ htmlWarning | renderMarkdown | safe }}</div>
+    {%- if tabWarning -%}
+      <div class="govuk-warning-text app-example__warning">
+        <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+        <strong class="govuk-warning-text__text">
+          <span class="govuk-visually-hidden">Warning</span>
+          {{ tabWarning | renderMarkdown | safe }}
+        </strong>
+      </div>
```
chrispymm: I thought the decision was that we weren't doing this today? But that we were going to wait and see if it was needed? Or have there been further conversations I'm not aware of?

PR #2038, opened 2025-11-20
docs(contribution): Timeout warning

chrispymm:
The component provides a modal dialog with a timeout warning to inform users when they have a limited amount of time to complete their current activity.  The modal contains a button which allows users to extend the time they have to complete the task.  This is important because whenever a process has a time restriction of less than 20 hours to complete a process users must be given a way to turn off, adjust, or extend the time limit in order to satisfy WCAG SC 2.2.1 Timing Adjustable (Level A).

After a configurable amount of time without user interaction, the timeout warning modal will show. This will display a countdown of the time remaining.  The modal contains a button that when pressed will make a request to a URL which will extend the length of their session (this behaviour needs to be developed in the backend of the application).  If no action is taken before the end of the countdown, the user will be redirected to another page.  This page should explain to the user that their session has been ended to keep their data secure, and that they will need to restart the process.

The component works with screenreaders, announcing the time remaining every 15 seconds up until the final 20 seconds which will be announced every second.

### DIFF
--- a/submissions/submission-1763680654726/code/example-1.js
+++ b/submissions/submission-1763680654726/code/example-1.js
@@ -120,7 +120,6 @@ TimeoutWarning.prototype.startUiCountdown = function () {
   var seconds = 60 * minutes
 
   $countdown.innerText = minutes + ' minute' + (minutes > 1 ? 's' : '')
-
   ;(function runTimer() {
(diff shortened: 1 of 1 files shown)


helennickols:
lede: Use the timeout warning component to let users know that they have a limited time to complete their task.

Issue #2111, opened 2025-12-22
Update 'To be reviewed' status explanation


chrispymm:
## What

Update the content for the 'To be reviewed' status

## Why

The current content says 'hasn't been recently updated'. and this may not always be true.  We have the situation where we need to make accessiblity fixes to the scrollable pane component.  It will have been updated, but will still need to be reviewed.

## Who needs to be involved

- Content Designer

## Done

When the content has been updated on the site.

PR #2119, opened 2026-01-07
docs: change for TBR to cover a11y reviews

helennickols:
A change as part of this ticket. 

https://github.com/ministryofjustice/moj-frontend/issues/2111


timminor (approved):
Loving your work! :)

PR #2128, opened 2026-01-15
feat!(v9.0.0): update to support GOV.UK Frontend v6.0.0

chrispymm:
BREAKING CHANGES: Updates to MOJ Frontend to support all the changes in GOV.UK Frontend v6.0.0

### DIFF
--- a/.github/workflows/sass.yml
+++ b/.github/workflows/sass.yml
@@ -16,11 +16,8 @@ jobs:
 
       matrix:
         package-version:
-          # Release without major deprecations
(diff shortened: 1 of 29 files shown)


Johan-D-MOJ (approved):
my only concern was about the fact that we are pulling a beta library. But this package is also marked as beta.
This is not a mandatory change but rather a suggestion so that that this beta version is not the default package for fresh install. I don't know how disruptive beta versions for GOVUK could be.

PR #2150, opened 2026-01-21
docs: adding unique and accessible titles for iframes

helennickols:


docs/components/date-picker/examples/excluded-days/index.njk
```diff
@@ -1,6 +1,6 @@
 ---
 layout: layouts/example.njk
-title: Date Picker Excluded Days (example)
+title: Date picker with excluded dates (example)
```
chrispymm:
```suggestion
title: Date picker with excluded days (example)
```

This is the same as the previous file. It might not actually get used though

PR #2178, opened 2026-02-13
updating the colour for the hover state of the opened menu to increas…

Johan-D-MOJ:
The contrast for the hover state of the open menu for the header is not high enough

Identify the bug
It was flagged up during accessibility assesment.

Description of the change
This change introduced a hover state for the opened menu of the header for the PDS-header component. And to avoid duplication, it introduced a shared mixin to use with the focus-visible state.

Possible drawbacks
failing css generation

Verification process
I ran `npm run build:package` with the change

Release notes
Not applicable


chrispymm (approved):
All good to go! 🚀

PR #2224, opened 2026-03-25
chore(renovate): update renovate config to match new repo labels

chrispymm:
After updating all the labels for the repo, we need to adjust the renovate config to apply the right label to new PRs, and respond to the correct stop updating label


Johan-D-MOJ (approved):
simple enough - let it use the new label!

PR #2271, opened 2026-04-27
build: replace gulp-immagemin with sharp

chrispymm:
The `gulp-imagemin` package has dependencies that are old and have security issues.  

The package itself is no longer actively maintained and a lot of its plugins are abandoned.

This PR replaces gulp-imagemin with the 'sharp' image transformation library - this is widely used and under active development which is much better from a security standpoint.

N.B. We have no gif images, and are unlikely to in the future so this new pipeline doesn't handle them.


gulp/build-compress-images.js
```diff
@@ -1,20 +1,47 @@
...
+
+function compressImages() {
+  return new Transform({
+    objectMode: true,
+    async transform(file, _, callback) {
+      if (file.isNull()) return callback(null, file)
+
+      const ext = file.extname.toLowerCase()
+
+      if (ext === '.svg') {
```
Johan-D-MOJ: this could be added to the try/catch. so SVG errors can also be caught and dealt with properly.

gulp/build-compress-images.js
```diff
@@ -1,20 +1,47 @@
...
+          pipeline = pipeline.jpeg({ quality: 80 })
+        }
 
-// Compresses png, jpeg, gif, png and svg images
-gulp.task('build:compress-images', async () => {
-  const { default: imagemin } = await import('gulp-imagemin')
+        file.contents = await pipeline.toBuffer()
+        return callback(null, file)
+      } catch (err) {
+        return callback(err)
```
Johan-D-MOJ, line 34: can we add a console.err statement here so we can see the error before it is raised up.

Johan-D-MOJ (changes requested):

Johan-D-MOJ (approved):

PR #2348, opened 2026-05-19
docs: new entries

helennickols:


chrispymm:
For some reason the example text for 'no results' is showing up in bold - so it looks like a heading, I'm assuming this is unintentional?
image

chrispymm:
Sorry - was looking at the wrong markdown file! I can see that it was intentional, but I don't think this should be an H4 - it is not actually providing a heading to the page content.  If we simply want visual differentiation it should probably just be bolded using ** syntax

chrispymm (approved):

PR #2354, opened 2026-05-20
fix: Offering to add a class name to the PDS header component

### LINKED ISSUE #43: Add containerClasses option for header
## What

Add a containerClasses input option to the header Nunjucks macro

## Why

This is to allow prototype builders to add CSS classes to the header where necessary. For instance to adjust the width of the header's content to their content width.

## Who needs to be involved

- Developer

## Done

CSS classes can be added to the header's container using the Nunjucks macro

Johan-D-MOJ:
Description of the change
This PR introduces the className attribute to PDS components.
The component has been updated to accept and render a className parameter.

### DIFF
--- a/docs/probation/components/pds-header/_arguments.md
+++ b/docs/probation/components/pds-header/_arguments.md
@@ -4,6 +4,7 @@
 | ----------------- | ------ | -------- | --------------------------------------------------------------------------------------------- |
 | environmentName | string | No      | Tag label showing the current environment. Hidden if value is `PRODUCTION`. |
 | environmentNameColour      | string | No      | Custom colour for the environment tag, overriding the default.                |
+| classname | string | No | Optional class name to apply to the main header element.                |
 | name        | string  | Yes       | The display name of the user.                           |
 | manageDetailsLink  | string | Yes       | URL linking to the user account management page.   |
 | servicesLink  | string | No | URL linking to a services page when JavaScript is disabled |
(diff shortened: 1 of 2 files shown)


docs/probation/components/pds-header/_arguments.md
```diff
@@ -4,6 +4,7 @@
 | ----------------- | ------ | -------- | --------------------------------------------------------------------------------------------- |
 | environmentName | string | No      | Tag label showing the current environment. Hidden if value is `PRODUCTION`. |
 | environmentNameColour      | string | No      | Custom colour for the environment tag, overriding the default.                |
+| classname | string | No | Optional class name to apply to the main header element.                |
```
chrispymm: convention across govuk components and moj components is to uses a `classes` parameter name

chrispymm (changes requested):

chrispymm:
The PR title here should be a `fix: ` type as it's changing the api for a component nunjucks.  So it will need to be a patch release through semantic release

Johan-D-MOJ:
> The PR title here should be a `fix: ` type as it's changing the api for a component nunjucks. So it will need to be a patch release through semantic release

I forgot it, semantic-release should be happier now.

docs/probation/components/pds-header/_arguments.md
Johan-D-MOJ: it's been renamed.

chrispymm (approved):

PR #2370, opened 2026-06-02
docs: fix code block indentation on tabbed templates

chrispymm:
This PR fixes an issue with our example code blocks on tabbed documentation pages where the indentation is incorrect.  The whole block is indented too far, and the second line onwards is indented more than the first. This affects all code blocks, no matter what language.

image

The issue that was causing this was partly indentation issues within the `example.njk` file. But also an issue caused by the govuk-tabs macro - the [govuk tabs macro](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/tabs/template.njk) includes indentation in its macro e.g.

```
{{- _tabPanel(params, item, loop.index) | trim | indent(2, true) }}
```

Because the code examples are within a `<pre>` tag all whitespace is preserved meaning the code output retains these extra indents. 

This PR adds a general `dedent` filter which we apply to the whole output of the tabs macro. Another more specific macro `dedentGovUKTabsPanel` is then required to handle removing the indentation of the html within the tab panels.

**Result**
image
image

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install dependencies
         run: npm run setup
       - name: Install Playwright Browsers
-        run: npx playwright install --with-deps
+        run: npx playwright install chromium --with-deps
       - name: Build package
         run: ENV=test npm run build:package
       - name: Build docs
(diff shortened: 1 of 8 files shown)


njpreddy (approved):
LGTM

Johan-D-MOJ (approved):

PR #2382, opened 2026-06-08
docs: multi select ai

helennickols:


RoseBrandle72:
Adding the second line under 'When to use' has split the 2 paras talking about applying an action, and what the actions could be.
Screenshot 2026-06-11 at 17 03 58


I suggest moving the new line to the bottom of the section.

chrispymm (approved):

PR #2393, opened 2026-06-15
docs: only remove spaces from the start of a line if they are whitespace

chrispymm:
This PR fixes an issue with the dedent filters, where it could truncate html tags unintentionally:

image

Instead of `firstLine` `lastLine` boolean options.  The filters have been adjusted to only dedent the lines if the spaces to be removed are actually whitespace.  This is much more reliable and should prevent any malformed html.

image


RoseBrandle72:
I did a sampling of tabbed and non-tabbed component pages and the issue is fixed on those pages

njpreddy (approved):
LGTM

PR #2435, opened 2026-07-02
chore: update renovate config

chrispymm:
This PR updates the renovate config to ensure major version updates are not grouped. Grouping major updates together is problematic if one of the updates fails.  It is safer to handle major updates 1 by 1.  

This will lead to more PRs, but this config change also switches to autoapprove and automerge PRs for  minor or patch version updates (as long as all checks pass). This should mean that the only renovate PRs that  need manual intervention are minor and patch updates that cause a failure, and major versions.

N.B. The change to the 'Production' label is to invalidate renovates cache so that the old PR's get removed and new ones created.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -12,6 +12,7 @@
     },
     {
       "groupName": "Build",
+      "matchUpdateTypes": ["minor", "patch"],
       "matchPackageNames": [
         "@babel/*",
         "@rollup/*",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -81,7 +87,19 @@
     },
     {
       "groupName": "Tools",
+      "matchUpdateTypes": ["minor", "patch"],
       "matchPackageNames": ["concurrently", "gulp", "gulp-*", "nodemon"]
+    },
+    {
+      "matchUpdateTypes": ["minor", "patch"],
+      "autoApprove": true,
```
Johan-D-MOJ, line 95: interesting to watch the behaviour in the next couple of run...

Johan-D-MOJ (approved):

renovate.json
chrispymm: 🤞

PR #2454, opened 2026-07-03
chore: add schedule to renovate config and remove dependabot config

chrispymm:
- Add explicit github action manager in renovate config
- Add a schedule to renovate
- Remove github action config from dependabot, as we don't need two tools generating the same PR. N.B. didn't remove the dependabot.yml file in case there are org checks that we have one :)


njpreddy (approved):
LGTM

PR #2471, opened 2026-07-06
docs: verify csrf token before multer file upload

chrispymm:
The CSRF middleware is refactored to accept tokens from both the query string and the request body, improving flexibility and security. 

This change is made because the previous csrf check pulled the token from `req.body` however the express middleware that puts the form data into `req.body` doesn't run on multipart form submissions required for uploads (this is why we need multer). Multer handles the request and creates `req.body`. However this means we cannot check the CSRF token before multer runs 😢. 

To ensure we can check the csrf token before multer we have to pass it as a query parameter on the form action.  So the `verifyCSRF` method has been updated to handle csrf tokens both in the request body and in a query parameter.

This change then required updating the session handling for form data to use `req.path` instead of `req.url`. This is because `req.url` contains the whole url including the query string (e.g /component-image?_csrf=lklkdfjl) and `req.path` just contains the path (e.g. /component-image). This change ensures more reliable session keying that doesn't break if query strings are added.


chrispymm:
Successfully completed a component submission on the preview site.

✅ PR Raised
✅ Issue Raised
✅ Submitter thanks email sent
✅ Design system team email sent

njpreddy (dismissed):
LGTM

PR #2503, opened 2026-07-20
fix: rename to params for classes properly.

Johan-D-MOJ:
Renaming the classes parameter in HMPPS header nunjucks file so it picks up the parameters properly.


njpreddy (approved):
LGTM

PR #2530, opened 2026-07-29
feat(search)!: update to official

### LINKED ISSUE #1382: Search component accessibility fix
## What

Document the accessibility issue for the components on the site, resolve the identified issue, implement the update, and notify users accordingly. This ticket expand from [#1110](https://github.com/ministryofjustice/moj-frontend/issues/1110).

**Statement from the Uservision report**

> When users type text into the search field in the page header, an ‘x’ button appears within the input field. Clicking this button deletes the input text. However, the button is not accessible via the keyboard, preventing keyboard-only users from clearing the input easily.
> 
> Furthermore, the mouse cursor does not change when hovering over the button. This lack of a focus indicator or hover state can confuse users who rely on visual cues to identify interactive elements.
> 
> The implementation appears to use JavaScript or custom CSS to insert the ‘x’ button dynamically when the search input receives focus, as the icon does not exist in the static HTML code.

## Why

We need to fix this as it is a WCAG fail

## Who needs to be involved

- Developer
- Accessibility Specialist

## Done

When the fix ensures the component will meet WCAG criteria

chrispymm:
Updates to the search component to bring it up to official status

* uses `<search>` element
* adds ability to add csrf token to form
* fixes #1382 by hiding non-keyboard-focusable browser added 'x' button
* adds icon button option
* adds stacked layout variant


njpreddy (dismissed):
LGTM

Johan-D-MOJ (approved):

PR #2544, opened 2026-08-17
feat(add another): add id attribute for component

### LINKED ISSUE #2548: MOJ add another - changes to addAnother button
There's an issue where this component only has one value for the addanother button. It doesn't have a submit button value. This makes it harder to implement no JS fallback. It also doesn't have an ID

chrispymm:
If an `id` option is passed to the component then it is added to the root element.  The id  will also be used to add `id`, `name` and `value` attributes to the add another button.  This enables the button to be identified by the server and allows the button to be targeted by error messages when implementing a no-javascript fallback.

The following nunjucks code

```njk
{{ mojAddAnother({
    id: "add-participants",
    itemLabel: "Participant",
    items: [
      ...
    ]
  }) }}
```

would result in the following html

```html
<div id="add-participants" class="moj-add-another" data-module="moj-add-another" data-item-label="Participant" data-moj-add-another-init="">
  <div class="moj-add-another__items">
    ...
  </div>
  <div class="moj-add-another__add-button-container">
    <button type="button"
            id="add-participants-add-another-button"
            class="govuk-button govuk-button--secondary moj-add-another__add-button"
            name="add-participants-add-another" 
            value="add-another-participant" 
            data-module="govuk-button"        
            data-govuk-button-init="">
      Add another participant
    </button>
  </div>
</div>
```

### DIFF
--- a/src/moj/components/add-another/add-another.nunjucks.spec.mjs
+++ b/src/moj/components/add-another/add-another.nunjucks.spec.mjs
@@ -85,6 +85,24 @@ describe('add another', () => {
       expect(addButton).toBeInTheDocument()
     })
 
+    test('adds id to the component and derived attributes to the add button', () => {
+      expect($component).toHaveAttribute('id', 'default-add-another')
+
+      const addButton = getByRole($component, 'button', {
+        name: 'Add another person'
+      })
+
+      expect(addButton).toHaveAttribute(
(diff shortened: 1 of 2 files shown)


Johan-D-MOJ (approved):
Looks good to me

PR #2555, opened 2026-08-19
docs: use an moj warning alert for pds components tab warning

chrispymm:
Users were not noticing/reading the warning about using the PDS components code examples in production. 

This PR updates the warning paragraph to use an MOJ Alert component in order to draw more attention to it.

**Before:**
image

**After:**
image


njpreddy (approved):
Looks good to me.

